PR #1, opened 2017-12-12
Fixed: System.IndexOutOfRangeException when using parameters

tobi4s:
Issue: "Console.exe sync -f" causes an IndexOutOfRangeException
Workaround: Use "Console.exe sync -f -f"
Fixed: Counter issue at line 1135 "i+1" should be "i"


kspearrin:
Thanks @tobi4s. This application was completely re-written so this PR is no longer applicable.

PR #9, opened 2018-10-22
Add support for filtering users based on their group membership

hajekj:
As discussed with the support e-mail, I am creating a pull request to allow filtering users coming from Azure AD based on their group membership.

Basically, back-compatibility with existing filtering is still kept which means that tenants who use the `include:` and `exclude:` keywords will still work fine. I added two keywords `includeGroup:` and `excludeGroup:` which should contain the group ID from Azure Active Directory. With each member found, the Graph endpoint [checkMemberGroups](https://developer.microsoft.com/en-us/graph/docs/api-reference/v1.0/api/user_checkmembergroups) is called and based on the query result the user is included or excluded.

This PR allows for better synchronization use in environments which have a lot of users including service accounts, B2B guest users etc. in Azure AD who shouldn't be synchronized into Bitwarden's organization.

### DIFF
--- a/src/services/azure-directory.service.ts
+++ b/src/services/azure-directory.service.ts
@@ -95,7 +95,7 @@ export class AzureDirectoryService extends BaseDirectoryService implements Direc
                         continue;
                     }
                     const entry = this.buildUser(user);
-                    if (this.filterOutUserResult(setFilter, user)) {
+                    if (await this.filterOutUserResult(setFilter, user)) {
                         continue;
                     }
 
@@ -144,9 +144,9 @@ export class AzureDirectoryService extends BaseDirectoryService implements Direc
             userSetType = UserSetType.IncludeUser;
         } else if (keyword === 'exclude') {
             userSetType = UserSetType.ExcludeUser;
-        } else if (keyword === 'includeGroup') {
+        } else if (keyword === 'includegroup') {
             userSetType = UserSetType.IncludeGroup;
-        } else if (keyword === 'excludeGroup') {
+        } else if (keyword === 'excludegroup') {
             userSetType = UserSetType.ExcludeGroup;
         } else {
             return null;
--- a/src/services/azure-directory.service.ts
+++ b/src/services/azure-directory.service.ts
@@ -178,16 +178,20 @@ export class AzureDirectoryService extends BaseDirectoryService implements Direc
                     return false;
                 }
             } else {
-                let memberGroups = await this.client.api(`/users/${user.id}/checkMemberGroups`).post({
-                    groupIds: Array.from(setFilter[1])
-                });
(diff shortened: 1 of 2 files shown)


hajekj:
Just a heads-up: I have just tested it with our Bitwarden deployment and it works just as expected.

src/services/azure-directory.service.ts
```diff
@@ -123,6 +123,83 @@ export class AzureDirectoryService extends BaseDirectoryService implements Direc
...
+        if (setFilter != null) {
+            let userSetTypeExclude = null;
+            if (setFilter[0] === UserSetType.IncludeUser) {
+                userSetTypeExclude = false;
+            } else if (setFilter[0] === UserSetType.ExcludeUser) {
+                userSetTypeExclude = true;
+            }
+            if (userSetTypeExclude != null) {
+                const entry = this.buildUser(user);
+                if (this.filterOutResult([userSetTypeExclude, setFilter[1]], entry.email)) {
```
kspearrin: Just `return this.filterOutResult([userSetTypeExclude, setFilter[1]], entry.email)`

src/services/azure-directory.service.ts
```diff
@@ -123,6 +123,83 @@ export class AzureDirectoryService extends BaseDirectoryService implements Direc
...
+    private async filterOutUserResult(setFilter: [UserSetType, Set<string>], user: graphType.User): Promise<boolean> {
+        if (setFilter != null) {
+            let userSetTypeExclude = null;
+            if (setFilter[0] === UserSetType.IncludeUser) {
+                userSetTypeExclude = false;
+            } else if (setFilter[0] === UserSetType.ExcludeUser) {
+                userSetTypeExclude = true;
+            }
+            if (userSetTypeExclude != null) {
+                const entry = this.buildUser(user);
```
kspearrin: Shouldn't need to build this again. Pass the entry in instead of `graphType.User`. The id should still be available under `entry.referenceId` or `entry.externalId` for your `checkMemberGroups` API call.

src/services/azure-directory.service.ts
hajekj: I have pushed the fix to the repo.

src/services/azure-directory.service.ts
hajekj: I have pushed the fix to the repo.

kspearrin:
Looks good, but please revert the changes to package-lock.json

hajekj:
Just reverted the changes in that file. Accidentally commited it, sorry.

hajekj:
Thanks for merging this. Would you like me to update the docs as well to include the changes?

kspearrin:
> 
> 
> Thanks for merging this. Would you like me to update the docs as well to include the changes?

That would be great.

PR #11, opened 2019-01-21
Directory synchronization improvements for Azure AD

hajekj:
As per https://community.bitwarden.com/t/directory-synchronization-improvements-for-azure-ad/3968

These changes introduce more efficient user synchronization based on group membership. The synchronization for `includeGroup:` selector is now based on delta query, which is more efficient than doing delta query on all users and then filtering them based on membership.

- Aditionally bumped version to 2.5.0
- Also fixed the issue with groups which had name longer than 50 characters for the sync client to trim it

### DIFF
--- a/src/services/azure-directory.service.ts
+++ b/src/services/azure-directory.service.ts
@@ -306,15 +306,23 @@ export class AzureDirectoryService extends BaseDirectoryService implements Direc
         entry.name = group.displayName;
 
         const memReq = this.client.api('/groups/' + group.id + '/members');
-        const memRes = await memReq.get();
-        const members: any = memRes.value;
-        if (members != null) {
-            for (const member of members) {
-                if (member[ObjectType] === '#microsoft.graph.group') {
-                    entry.groupMemberReferenceIds.add((member as graphType.Group).id);
-                } else if (member[ObjectType] === '#microsoft.graph.user') {
-                    entry.userMemberExternalIds.add((member as graphType.User).id);
+        let memRes = await memReq.get();
+        while (true) {
+            const members: any = memRes.value;
+            if (members != null) {
+                for (const member of members) {
+                    if (member[ObjectType] === '#microsoft.graph.group') {
+                        entry.groupMemberReferenceIds.add((member as graphType.Group).id);
+                    } else if (member[ObjectType] === '#microsoft.graph.user') {
+                        entry.userMemberExternalIds.add((member as graphType.User).id);
+                    }
                 }
+            }
+            if (memRes[NextLink] == null) {
+                break;
+            } else {
+                const nextMemReq = this.client.api(memRes[NextLink]);
+                memRes = await nextMemReq.get();
             }
         }
 


kspearrin:
Thanks for the PR here. I will have to make some time to review this ore in-depth.

hajekj:
No problem. So far been using it successfuly in our directory-connector fork against your hosted instance and seems to work fine.

The logic is quite simple - we check whether we are to include a group or not - if not, regular `/users/delta` endpoint is used. If yes, we filter the specific groups that we want to include `/groups/delta?$filter=id eq 'groupId' or ...&$select=members` which lists the group members (only object ids tho, the expand is not working for this endpoint). We then fetch the full user from the Graph and use their info to provision them and match them to groups.

hajekj:
@kspearrin any update on this one? I see there are some conflicts - the package.json is fine I guess, the azure-directory.service.ts can be easily merged...

kspearrin:
We did some updates recently to the Azure AD service that got rid of the delta querying altogether due to its limitations in determining user group changes, so I am not sure how this affects that you are trying to achieve here.

hajekj:
Using deltas for detecting group membership changes makes sense in my opinion. The thing is, that since you removed the deltas, the sync has no way to detect users which have been removed from AAD (deleted) and therefor is quite a breaking change.

Personally, I would stick to the deltas (unless you want to do some user comparison against users currently in Bitwarden's organization).

If you check this PR, it handles also deltas for group memberships as well, which is the desired behavior I would say. We have been using the proposed change in production for few months now without hitting any issues.

kspearrin:
> The thing is, that since you removed the deltas, the sync has no way to detect users which have been removed from AAD (deleted) and therefor is quite a breaking change.

Good point. Let me look into adding a delta query to look for deleted users.

hajekj:
Look at my PR please, it does all that.

________________________________

kspearrin:
From what I can tell, your change only introduces code to support a `IncludeGroup` filter and the original problem with using user delta queries remains.

kspearrin:
I added https://github.com/bitwarden/directory-connector/commit/b0247caa4c3ce16f75be0996f38938b4a5fb4366 to get deleted users via delta queries.

hajekj:
I think the code will need some more changes tho.

Right now, when a user is removed from a group specified with _includeGroup:_ they won't get removed from Bitwarden, because the code (https://github.com/bitwarden/directory-connector/blob/master/src/services/azure-directory.service.ts#L92) will just ignore them.

The original PR I think was well detecting the user group changes (via group/groupid/delta query) which works fine.

What we can do, is either go back to this code (that's what I would go with) or figure out some way to modify filterOutUserResult - which will result in having to send non-existant users to the BW API for delete as well (because the connector isn't aware of users provisioned to BW).

kspearrin:
The problem is that delta queries with Azure AD do not detect a user's group reassignment, which causes a bug. This is why the current solution was implemented. I am not sure how to solve this fundamental problem while still using a delta query.

hajekj:
Maybe I am thinking all wrong about this, but you shouldn't need users to know group membership changes?

Because you run getGroups on every sync run (https://github.com/bitwarden/directory-connector/blob/master/src/services/azure-directory.service.ts#L268) which results in all groups including memberships being pulled (https://github.com/bitwarden/directory-connector/blob/master/src/services/azure-directory.service.ts#L301) from Graph.

Now shouldn't it be enough to just send the groups + members to your API which pairs it to existing Bitwarden users by their externalId?

kspearrin:
@hajekj The latest version of directory connector (and server) now support an "Overwrite existing organization users based on current sync settings.". This setting always performs a full (non-delta) sync and replaces the organization's existing directory users/group memberships with the newly sync data set. I think this gets rid of the issue of missing group membership removals that we previously mentioned. This might become the new default in a future update.

hajekj:
I quickly checked it and I like it. Also, I noticed that when you are pulling users (https://github.com/bitwarden/directory-connector/blob/master/src/services/azure-directory.service.ts#L81) you are not including the `accountEnabled` property. The request should contain `?$select=accountEnabled,displayName,mail,userPrincipalName,id,givenName,surname` to select the properties and include the accountEnabled one.

Ref: https://developer.microsoft.com/en-us/graph/graph-explorer

https://github.com/bitwarden/directory-connector/blob/79cedfcb7b6948b9ad4d7aef6774cfc09480a20b/src/services/azure-directory.service.ts#L117

Issue #6, opened 2018-06-05
Support for STARTTLS with LDAP


Techwolf12:
As send per mail:

For STARTTLS you first have to connect to a unencrypted connection, then you have to start your TLS session with client.starttls: http://ldapjs.org/client.html#starttls

You should also catch an error if it occurs, my suggestion is to make a list instead of checkbox with:

Encryption: None, STARTTLS, SSL

PR #33, opened 2020-03-01
starttls: Support LDAP STARTTLS

colin-campbell:
Fixes bitwarden/directory-connector#6
Enables support for STARTTLS on plain LDAP connections, with some UI changes to accommodate both encrypted connection types.

STARTTLS. 

image

LDAPS. 

image

### DIFF
--- a/src/models/ldapConfiguration.ts
+++ b/src/models/ldapConfiguration.ts
@@ -1,6 +1,8 @@
 export class LdapConfiguration {
-    ssl = false;
-    sslAllowUnauthorized = false;
+    encrypted = false;
+    encryptionType: string = 'starttls';
+    tlsCaPath: string;
+    certDoNotVerify = false;
     sslCertPath: string;
     sslKeyPath: string;
     sslCaPath: string;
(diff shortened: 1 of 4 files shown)


src/models/ldapConfiguration.ts
```diff
@@ -1,6 +1,8 @@
 export class LdapConfiguration {
-    ssl = false;
```
kspearrin: This is a breaking change for older versions. Their configs will load with this data.

colin-campbell: Thank you. I will re-roll to accommodate older configs.

colin-campbell: Have restored the original config variables, so that old configs should load as before.

kspearrin:
I see you had starttls option defaulting to true. Any reason why? Shouldn't it be false as a default?

colin-campbell:
In my experience, starttls is by far the most common configuration historically with these older protocols such as SMTP and LDAP. Many instances use it for compatibility purposes. But of course feel free to change the default.

PR #44, opened 2020-05-19
Enable filtering Azure AD directory by administrative unit

psiniemi:
Administrative units are a new Azure AD feature where you can add any number of administrative units to a group or other AAD object. It is a convenient way for us select just the groups we want in bitwarden by tagging them with the "Bitwarden" administrative unit.


src/services/azure-directory.service.ts
```diff
@@ -170,6 +170,54 @@ export class AzureDirectoryService extends BaseDirectoryService implements Direc
         return entries;
     }
 
+    protected async createAADCustomSet(filter: string): Promise<[boolean, Set<string>]> {
```
cscharf, line 173: This may be a `private` (vs. `protected`) method, it's not needed for inheritance.
For naming conventions, please rename this method, `createAadCustomSet`.

cscharf (commented):
Hi @psiniemi , thank you for your contribution! just one small naming convention nit but otherwise looks good.

cscharf (approved):
Awesome, thanks!

cscharf:
@fschillingeriv , this will enhance the directory connector for Azure syncs with some new capabilities, noted on the help site here: https://bitwarden.com/help/article/user-group-filters/#azure-active-directory

Adding the filter expressions, `excludeadministrativeunit:<name>` and `includeadministrativeunit:<name>` relating to [Azure AD Administrator Units (AUs)](https://docs.microsoft.com/en-us/azure/active-directory/roles/administrative-units). This would work exactly how the Group filters work today, except it would use an Azure AU instead of a Group.

Issue #40, opened 2020-04-29
Missing paged control


AdamVB:
This Issue is similar to the one that has already been closed: https://github.com/bitwarden/directory-connector/issues/16

Essentially we get the same error message. But the proposed solution (remove repeated pathing) didn't work out for us.

Here's our config:

```
{

  "directoryType": 0,
  "directoryConfig_0": {
    "ssl": true,
    "sslAllowUnauthorized": true,
    "port": "636",
    "currentUser": false,
    "ad": false,
    "hostname": "ldap.otc-service.internal",
    "rootPath": "dc=otc,dc=app",
    "username": "cn=bind_username,ou=profile,dc=otc,dc=app",
        "sslCaPath": "xxxxxxxx.crt",
    "password": "xxxxxxx"
  },
  "directoryConfig_2": {},
  "directoryConfig_1": {},
  "directoryConfig_3": {},
  "syncConfig": {
    "users": true,
    "groups": false,
    "interval": 5,
    "removeDisabled": false,
    "overwriteExisting": false,
    "useEmailPrefixSuffix": false,
    "creationDateAttribute": "whenCreated",
    "revisionDateAttribute": "whenChanged",
    "emailPrefixAttribute": "sAMAccountName",
    "memberAttribute": "memberUid",
    "userObjectClass": "person",
    "groupObjectClass": "posixGroup",
    "userEmailAttribute": "mail",
    "groupNameAttribute": "cn",
    "userFilter": "",
    "groupFilter": "",
    "groupPath": "ou=group",
    "userPath": "ou=people"
  },
```
bind usernames are under ou=profile,,dc=otc,dc=app users are in ou=people,dc=otc,dc=app and groups in ou=group,dc=otc,dc=app

When doing the sync test it shows this error
```
./bwdc test
missing paged control
```

ldapsearch works fine with that bind user 

Any Ideas?

kspearrin:
@AdamVB Can you try debugging the LDAP quries that are generated?

> You can debug the LDAP queries from the output in the console under View > Developer Tools. Or, if using the CLI, you can set environment variable BWCLI_DEBUG=true to see the LDAP query debug output in terminal.

kareemshayeb:
@kspearrin using the CLI, the only message we get is this:

`missing paged control`

However, with the GUI tool we get this in the developer console:

[...]

PR #45, opened 2020-05-19
Added option to let user control the paged param

kspearrin:
Some LDAP servers have paging turned off. If so, this can cause an error since we made the assumption that all servers are using paging. This change lets the user control that through the settings. This can also be set in the data.json file under directoryConfig_0 > pagedSearch.

For example:

	"directoryConfig_0": {
		...,
		"pagedSearch": true
	},


kspearrin:
resolves #40

PR #46, opened 2020-06-16
include all members of the g suite group

kspearrin:
Users of a group can have roles of owner, manager, and member. Not sure why we were only allowing members to filter through before. We've had a few customers complaining about this. This PR allows all members to be part of the synced group.


kspearrin:
Correct.

PR #47, opened 2020-06-22
add support for email coming from username prop

kspearrin:
OneLogin users may not have a valid email address in the `email` field. According to our customers, a valid email address could also come from the `username` field. Finally, we also fall back to trying to form an email address using the `useEmailPrefixSuffix` sync setting, if enabled.

### DIFF
--- a/src/services/onelogin-directory.service.ts
+++ b/src/services/onelogin-directory.service.ts
@@ -12,6 +12,9 @@ import { DirectoryService } from './directory.service';
 import { I18nService } from 'jslib/abstractions/i18n.service';
 import { LogService } from 'jslib/abstractions/log.service';
 
+// Basic email validation: something@somthing.something
+const ValidEmailRegex = /\S+@\S+\.\S+/;
+
 export class OneLoginDirectoryService extends BaseDirectoryService implements DirectoryService {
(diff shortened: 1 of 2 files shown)


src/services/onelogin-directory.service.ts
```diff
@@ -82,9 +85,23 @@ export class OneLoginDirectoryService extends BaseDirectoryService implements Di
         const entry = new UserEntry();
         entry.externalId = user.id;
         entry.referenceId = user.id;
-        entry.email = user.email != null ? user.email.trim().toLowerCase() : null;
         entry.deleted = false;
         entry.disabled = user.status === 2;
+        entry.email = user.email != null ? user.email : null;
+        const emailInvalid = (ue: UserEntry) => ue.email == null || ue.email === '';
+        if (emailInvalid(entry) && user.username != null && user.username !== '') {
```
cscharf, line 92: could be simplified to `if (emailInvalid(entry) && !!user.username) {`

kspearrin: I prefer to be a bit more explicit in JS and try to avoid truthy/undefined conditions as much as possible since they are not directly translatable to C#

PR #51, opened 2020-07-02
fix okta paging

kspearrin:
Okta SDK does not perform paging properly. Get rid of it and just call APIs directly.

ref: https://github.com/okta/okta-sdk-nodejs/issues/77

Also add some throttling to group membership API calls to avoid Okta's rate limit.


package.json
```diff
@@ -199,9 +199,10 @@
     "@angular/router": "7.2.1",
     "@angular/upgrade": "7.2.1",
     "@microsoft/microsoft-graph-client": "1.2.0",
-    "@okta/okta-sdk-nodejs": "2.0.1",
+    "@types/bent": "^7.0.2",
     "angular2-toaster": "6.1.0",
     "angulartics2": "6.3.0",
+    "bent": "git+https://git@github.com/kspearrin/bent.git",
```
kspearrin, line 205: Switched to `bent` package for doing http in node. I had to compile my own version of bent so that it would use the nodejs implementation rather than it's browser implementation which uses `fetch`. `fetch` will not return the `link` response header in CORS situations. This is actually the reason why the okta library doesn't work with paging. It uses fetch.

Issue #55, opened 2020-07-31
Dynamic Google Groups aren't populated


infraweavers:
If you have a group that contains (for example) the whole company i.e. "All users in the organisation":
![image](https://user-images.githubusercontent.com/36475663/89022374-d803eb00-d319-11ea-961f-bc2ca3a08789.png)

then the bitwarden sync does not put all the users in the group when being sync'd

infraweavers:
Believe this is fixed in: https://github.com/bitwarden/directory-connector/pull/56

infraweavers:
The only method I'm aware of is "Add all current and future users of company to this group" : 
![image](https://user-images.githubusercontent.com/36475663/89026927-73e52500-d321-11ea-9bf7-bdd83abb2dd6.png)

infraweavers:
Our actual usecase is that we'd like to use this group to grant everyone access to our company-wide shared passwords automatically, then have 2 other vaults with only specific groups have access. It seems like the only way to implement this without having to manually reassign users into groups within bitwarden would be to have this "all users" group working

PR #56, opened 2020-07-31
Fix for #55

infraweavers:
This should ask the gSuite API to enumerate all users in dynamic groups etc: https://developers.google.com/admin-sdk/directory/v1/reference/members/list#request

### DIFF
--- a/src/services/gsuite-directory.service.ts
+++ b/src/services/gsuite-directory.service.ts
@@ -49,7 +49,7 @@ export class GSuiteDirectoryService extends BaseDirectoryService implements Dire
 
         await this.auth();
 
-        let users: UserEntry[];
+        let users: UserEntry[] = [];
         if (this.syncConfig.users) {
             users = await this.getUsers();
         }
--- a/src/services/gsuite-directory.service.ts
+++ b/src/services/gsuite-directory.service.ts
@@ -194,14 +194,14 @@ export class GSuiteDirectoryService extends BaseDirectoryService implements Dire
                     }
                     const type = member.type.toLowerCase();
                     if (type === 'user') {
-                        if (member.status.toLowerCase() !== 'active') {
+                        if (member.status == null || member.status.toLowerCase() !== 'active') {
                             continue;
                         }
                         entry.userMemberExternalIds.add(member.id);
                     } else if (type === 'group') {
                         entry.groupMemberReferenceIds.add(member.id);
                     } else if (type === 'customer') {
-                        for(let user of users) {
+                        for (let user of users) {
                             entry.userMemberExternalIds.add(user.externalId);
                         }
                     }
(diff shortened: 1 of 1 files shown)


infraweavers:
@kspearrin does this look good to solve https://github.com/bitwarden/directory-connector/issues/55 ?

kspearrin:
Yes, it appears to solve that use case, however, I am worried that this brings in a breaking change for existing users.

infraweavers:
@kspearrin There is an alternative way of doing it, however it will rely on the undocumented "CUSTOMER" member of the group. Rather than asking the gsuite API to expand it, we can look for a member with type CUSTOMER; if there is one then add all users.

infraweavers:
For the record the response for one of these groups is:
```
{
  "kind": "admin#directory#members",
  "etag": "\"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA\"",
  "members": [
    {
      "kind": "admin#directory#member",
      "etag": "\"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA\"",
      "id": "000000000",
      "role": "MEMBER",
      "type": "CUSTOMER"
    }
  ]
}
```

infraweavers:
@kspearrin actually it's not undocumented, I was reading the wrong docs, it's here: https://developers.google.com/admin-sdk/directory/v1/reference/members#resource

infraweavers:
@kspearrin Reimplemented to do the adding of all users in the sync application rather than pushing that to the gsuite API; this should keep the change 100% backwards compatible.

src/services/gsuite-directory.service.ts
```diff
@@ -192,15 +192,18 @@ export class GSuiteDirectoryService extends BaseDirectoryService implements Dire
                     if (member.type == null) {
                         continue;
                     }
-                    if (member.status == null || member.status.toLowerCase() !== 'active') {
-                        continue;
-                    }
```
kspearrin, line 197: Why was this moved? Is this because a group of type customer has a different status than others? This should still allow to both user and group types, right?

infraweavers: Yes it does still work for both groups of groups and groups of users. The `status` property is the status of the member, i.e. whether the actual user is active or not. It isn't that you can have a disabled/inactive membership inside a group as Google Groups doesn't support that. So the status property is `null` for both the `member.type` of `group` and `member.type` of `customer` and only applies in the case of the `member.type` being `user`. So we've moved the check for member activity into the `user` block itself as otherwise it'll always continue.

Ref: https://developers.google.com/admin-sdk/directory/v1/reference/members#resource

PR #58, opened 2020-08-05
sso login from bwdc desktop app

kspearrin:
Support SSO login from desktop app

- Stub out SSO component
- Process deep link callback

This solution is incomplete. Needs the following still:

- Localization strings
- Set password flow

I'd also like to consider kicking the org identifier collection process over to the web vault. This would avoid some routing issues that currently exist on processing the callback.

### DIFF
--- a/src/app/accounts/sso.component.ts
+++ b/src/app/accounts/sso.component.ts
@@ -1,0 +1,41 @@
+import { Component } from '@angular/core';
+
+import {
+    ActivatedRoute,
+    Router,
+} from '@angular/router';
+
+import { ApiService } from 'jslib/abstractions/api.service';
+import { AuthService } from 'jslib/abstractions/auth.service';
+import { CryptoFunctionService } from 'jslib/abstractions/cryptoFunction.service';
+import { I18nService } from 'jslib/abstractions/i18n.service';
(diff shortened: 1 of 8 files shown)


src/app/accounts/sso.component.ts
```diff
@@ -0,0 +1,41 @@
...
+        super(authService, router, i18nService, route, storageService, stateService, platformUtilsService,
+            apiService, cryptoFunctionService, passwordGenerationService);
+        this.successRoute = '/tabs/dashboard';
+        this.redirectUri = 'bwdc://sso-callback';
+        this.clientId = 'connector';
+    }
+
+    async submit() {
+        await super.submit();
+        this.router.navigate(['login']);
```
kspearrin, line 39: Have to navigate away from the sso page, so that the callback can actually send us back to SSO page.

src/app/app.component.ts
```diff
@@ -133,6 +133,9 @@ export class AppComponent implements OnInit {
                             properties: { label: message.label },
                         });
                         break;
+                    case 'ssoCallback':
+                        this.router.navigate(['sso'], { queryParams: { code: message.code, state: message.state } });
```
kspearrin, line 137: If we are not navigated away from SSO page, this will fail and do nothing since we are already on SSO.

PR #93, opened 2021-02-03
dont catch api error and return false

kspearrin:
Customer was reporting that graph API calls were failing, which were being caught here and causing the user to not be filtered out of the synced result set. This allows the API call failure to propagate out further and return an error to the end user.

### DIFF
--- a/src/services/azure-directory.service.ts
+++ b/src/services/azure-directory.service.ts
@@ -272,20 +272,18 @@ export class AzureDirectoryService extends BaseDirectoryService implements Direc
             return this.filterOutResult([userSetTypeExclude, setFilter[1]], user.email);
         }
 
-        try {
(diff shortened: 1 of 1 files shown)


cscharf (approved):
The build is failing, but that looks like something else entirely, these changes look good.

cscharf:
@kspearrin , it looks like this Graph API call is now failing for most customers trying to sync deleted users with the following exception (after updating to the latest version):
```
POST: /users/{user.externalId}/checkMemberGroups
Request_ResourceNotFound : Resource '<{user.externalId}>' does not exist or one of its queried reference-property objects are not present
```
Trying to dig into root cause but having a very hard time with the poor documentation in GraphAPI around this or any examples/troubleshooting steps.

PR #99, opened 2021-03-05
Don't check user group filter for deleted users

cscharf:
## Overview
Apparently Azure AD gets angry when you try to check user groups for a deleted user and responds with the error:

```
Resource <deleted user ID> does not exist or one of its queried reference-property objects are not present
```

This took a while to figure out, of course, because when you look in AD at the deleted user, they are indeed still in that group, and if you restore the user, they are in the group, but if they're currently "deleted" and you try to run `checkUserGroups` through the graph API, you get an error.

Previously this check was contained within a `try {} catch {}` block, however that was removed due to other downstream failures causing issues (for different reasons) - see: https://github.com/bitwarden/directory-connector/pull/93.

Ultimately, the solution was to pass a flag to the filter method NOT to check user groups when we're in the context of querying only deleted users AND to leave off the `try {} catch {}` so other issues/errors still `throw` as expected.

### DIFF
--- a/src/services/azure-directory.service.ts
+++ b/src/services/azure-directory.service.ts
@@ -90,7 +90,7 @@ export class AzureDirectoryService extends BaseDirectoryService implements IDire
                         continue;
                     }
                     const entry = this.buildUser(user);
-                    if (await this.filterOutUserResult(setFilter, entry)) {
+                    if (await this.filterOutUserResult(setFilter, entry, true)) {
                         continue;
                     }
 
@@ -147,7 +147,7 @@ export class AzureDirectoryService extends BaseDirectoryService implements IDire
                     if (!entry.disabled && !entry.deleted) {
                         continue;
                     }
-                    if (await this.filterOutUserResult(setFilter, entry)) {
+                    if (await this.filterOutUserResult(setFilter, entry, false)) {
                         continue;
                     }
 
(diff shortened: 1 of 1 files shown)


src/services/azure-directory.service.ts
```diff
@@ -273,6 +274,10 @@ export class AzureDirectoryService extends BaseDirectoryService implements IDire
             return this.filterOutResult([userSetTypeExclude, setFilter[1]], user.email);
         }
 
+        // We need to *not* call the /checkMemberGroups method for deleted users, it will always fail
```
MGibson1, line 277: Only concern is: are deleted users always intended to be removed from all orgs?

I imagine yes, but thought I'd check

MGibson1: I guess even if the answer is no, what are we gonna do about it, right?

cscharf: Good question, before I explain I want to clarify language that the bwdc directory sync only applies to a single Org, whereas groups in the directory are sync'd to groups in the Org, and users to them, etc.

I _believe_ this is okay because:
* In this context an `"includegroup:..."` value has been applied to the `UserFilter`, which sets up a userSetType of `UserSetType.IncludeGroup`
* When doing this, we of course "filter" the user by doing a group membership check against the provide group IDs to see if the action is applicable
* This would normally be fine, if say the user was not deleted but only removed from the group, however that's not the case here
* This method is shared between getting active users and deleted users, there is no issue/error for active users, but in Azure apparently with deleted users, you can't call this check method, so that's the only time it's negated
* If a user is deleted in Azure, it's actually deleted (not just disabled or removed from the group, the user themselves are deleted), and therefore we can safely infer that they should be removed from the Org in Bitwarden (at least a safe assumption I think)

Hopefully that makes sense.

MGibson1: yup, thank you!

MGibson1 (approved):

PR #106, opened 2021-03-25
Fix filteringForUnsupportedUsers and extend email validation to handle emails up to 256 char

djsmith85:
While working on bitwarden/server#1101 I found another place where an email's length is checked.

I encountered a faulty behaviour though, while checking the method [filterUnsupportedUsers in syncService.ts](https://github.com/bitwarden/directory-connector/blob/61c6ba81894fb6162cd51d26eb542fb616cf7a29/src/services/sync.service.ts#L106) 

Also check [jsFiddle](https://jsfiddle.net/1t89mou6/37/) for test cases

Finally, I extended the filter to handle emails with up to 256 chars

### DIFF
--- a/src/services/sync.service.ts
+++ b/src/services/sync.service.ts
@@ -104,7 +104,7 @@ export class SyncService {
     }
 
     private filterUnsupportedUsers(users: UserEntry[]): UserEntry[] {
-        return users == null ? null : users.filter(u => u.email?.length <= 50);
+        return users == null ? null : users.filter(u => u.email?.length <= 256);
     }
 
     private flattenUsersToGroups(levelGroups: GroupEntry[], allGroups: GroupEntry[]): Set<string> {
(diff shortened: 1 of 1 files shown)


src/services/sync.service.ts
```diff
@@ -104,7 +104,7 @@ export class SyncService {
     }
 
     private filterUnsupportedUsers(users: UserEntry[]): UserEntry[] {
-        return users == null ? null : users.filter(u => u.email == null || u.email.length <= 50);
```
cscharf, line 107: 😮  wow. Good find!

cscharf (approved):

src/services/sync.service.ts
djsmith85: Thank you 😄  Yeah it took me awhile to figure out if filterUnsupportedUsers actually wanted to return unsupported users or filter them out because of the 'equals null' and the 'or'

PR #112, opened 2021-04-21
Resolve ExpressionChangedAfterItHasBeenCheckedError

Hinton:
## Objective
When clicking the start or sync buttons angular casts a `ExpressionChangedAfterItHasBeenCheckedError` error. This seems to be caused by the elements referring to themselves using the `appApiAction` directive. By wrapping the buttons in forms and attaching the event to the forms we avoid this error.

### DIFF
--- a/src/app/tabs/dashboard.component.html
+++ b/src/app/tabs/dashboard.component.html
@@ -14,21 +14,25 @@
             <strong *ngIf="syncRunning" class="text-success">{{'running' | i18n}}</strong>
             <strong *ngIf="!syncRunning" class="text-danger">{{'stopped' | i18n}}</strong>
         </p>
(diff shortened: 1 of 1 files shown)


eliykat (changes requested):
Looks good to me and also solves the issue on my machine.

However, the same changes need to be made for the "Test Now" button before we can merge.

I think we can also remove the `id` on the buttons now that they don't reference themselves. (minor nit)

I almost jumped in and made these changes, but I didn't know if that would be a faux pas 😅

Hinton:
@eliykat Good catch, I have nothing against anyone pushing fixes directly on my branch :) Especially with the time differences. There's also the handy suggestion comments which can easily be applied.

eliykat (approved):

eliykat:
Hitting merge on this one so we can get it across to QA before I check out for the night.

PR #116, opened 2021-05-03
Bump node to 14

Hinton:
## Objective
Follow up to #114 , bumps supported nodejs version from 10 to 14.

I also added `engines` to `package.json` to specify we only suport `node 14+` and `npm ~6`.

Depends on https://github.com/bitwarden/jslib/pull/358


package.json
```diff
@@ -202,24 +202,28 @@
...
-    "open": "7.1.0",
+    "open": "^8.0.6",
     "proper-lockfile": "^4.1.2",
     "rxjs": "^6.6.7",
     "tslib": "^2.2.0",
     "zone.js": "^0.11.4",
     "zxcvbn": "4.4.2"
+  },
+  "engines": {
+    "node": ">= 14",
```
MGibson1, line 226: Any particular reason not to specify 14.16.1 exactly? From what I'm reading this only creates warnings unless `"engineStrict"=true`

MGibson1 (commented):
Just the one question

package.json
Hinton: Changed to `~14`.

MGibson1 (dismissed):

MGibson1 (approved):

PR #117, opened 2021-05-13
Add node version to requirements in README.md

djsmith85:
With @Hinton bumping node to 14 recently, I've adjusted the README to show the need for Node 14.17 (LTS)

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ We provide detailed documentation and examples for using the Directory Connector
 
 **Requirements**
 
-- [Node.js](https://nodejs.org) v14.17 or greater
+- [Node.js](https://nodejs.org) v14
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -47,7 +47,7 @@ We provide detailed documentation and examples for using the Directory Connector
 
 **Requirements**
 
-- [Node.js](https://nodejs.org/)
+- [Node.js](https://nodejs.org) v14.17 or greater
```
eliykat, line 50: The requirement in package.json is `~14`  ([link](https://github.com/bitwarden/directory-connector/blob/master/package.json#L196)), which I understand to mean v14.0.0 and any subsequent patch releases.  So I think this should just say:

```suggestion
- [Node.js](https://nodejs.org) v14
```

eliykat (changes requested):

README.md
djsmith85: @eliykat Thank you for the review. I'd picked the LTS version as that would be most commonly used and also the one I added in the other repo's README. But no problem with letting it be v14, as that should also work without issues.

eliykat (approved):
Thanks!

PR #119, opened 2021-05-13
Add option to bypass large import limit of 2000 users

MGibson1:
# Overview

Related to bitwarden/server#1311. Allows import of large numbers of users and groups. Also add orgId to last sync hash.

# Files Changed
* **settings.component.html**: Add `largeImport` checkbox button. Formatting changes.
* **sync.command.ts**: remove unneccessary using
* **syncConfiguration.ts**: Add largeImport
* **sync.service.ts**: specify largeImport. Add orgId to last sync hash. This resolves issue where changing the organization synced to would not update until directory changes exist.

### DIFF
--- a/src/app/tabs/settings.component.html
+++ b/src/app/tabs/settings.component.html
@@ -283,8 +283,8 @@
                         <div class="form-check">
                             <input class="form-check-input" type="checkbox" id="useEmailPrefixSuffix"
                                 [(ngModel)]="sync.useEmailPrefixSuffix" name="UseEmailPrefixSuffix">
-                            <label class="form-check-label" for="useEmailPrefixSuffix">{{'useEmailPrefixSuffix' |
-                                i18n}}</label>
+                            <label class="form-check-label"
+                                for="useEmailPrefixSuffix">{{'useEmailPrefixSuffix' | i18n}}</label>
                         </div>
                     </div>
                     <div [hidden]="!sync.useEmailPrefixSuffix">
--- a/src/app/tabs/settings.component.html
+++ b/src/app/tabs/settings.component.html
@@ -37,8 +37,7 @@
                         <div class="form-check">
                             <input class="form-check-input" type="checkbox" id="pagedSearch"
                                 [(ngModel)]="ldap.pagedSearch" name="PagedSearch">
-                            <label class="form-check-label"
-                                for="pagedSearch">{{'ldapPagedResults' | i18n}}</label>
+                            <label class="form-check-label" for="pagedSearch">{{'ldapPagedResults' | i18n}}</label>
                         </div>
                     </div>
                     <div class="form-group">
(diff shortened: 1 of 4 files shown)


src/app/tabs/settings.component.html
```diff
@@ -277,8 +283,8 @@ <h3 class="card-header">{{'sync' | i18n}}</h3>
                         <div class="form-check">
                             <input class="form-check-input" type="checkbox" id="useEmailPrefixSuffix"
                                 [(ngModel)]="sync.useEmailPrefixSuffix" name="UseEmailPrefixSuffix">
-                            <label class="form-check-label"
-                                for="useEmailPrefixSuffix">{{'useEmailPrefixSuffix' | i18n}}</label>
+                            <label class="form-check-label" for="useEmailPrefixSuffix">{{'useEmailPrefixSuffix' |
```
cscharf, line 286: Seems like the formatting got a little wonky on these lines where they break after the `|` in the angular expression vs. breaking between HTML attributes (which _feels_ better).

src/services/sync.service.ts
```diff
@@ -67,22 +67,22 @@ export class SyncService {
...
             const reqJson = JSON.stringify(req);
 
+            const orgId = await this.configurationService.getOrganizationId();
+            if (orgId == null) {
+                throw new Error('Organization not set.');
+            }
+
             let hash: string = null;
-            const hashBuf = await this.cryptoFunctionService.hash(this.apiService.apiBaseUrl + reqJson, 'sha256');
+            const hashBuf = await this.cryptoFunctionService.hash(this.apiService.apiBaseUrl + orgId + reqJson, 'sha256');
```
cscharf, line 79: The first time this runs after updating, will it cause a forced full sync due to the hash being off/different than previously generated? Would or _could_ there be a transitionary way to do this (where we could create a tech-debt card in Asana to track it) where you get both hashes, compare either but store the new way? Then in a future update we can remove that old way code and only use the new way?

MGibson1: That's a really good idea.

What would happen is pretty cheap server side. We'd We would just send the sync despite knowing it contains no changes, the server would detect that all users/groups exist and contain no changes and just do nothing. It'd be potentially a lot of read operations (depending on the number of users/groups), but few writes, if any.

Still, we have this check for a reason and your solution allows a smooth transition -- nice

src/services/sync.service.ts
```diff
@@ -75,14 +75,20 @@ export class SyncService {
...
-            if (hashBuf != null) {
-                hash = Utils.fromBufferToB64(hashBuf);
+            const hashBuff = await this.cryptoFunctionService.hash(this.apiService.apiBaseUrl + orgId + reqJson, 'sha256');
+            if (hashBuff != null) {
+                hash = Utils.fromBufferToB64(hashBuff);
             }
             const lastHash = await this.configurationService.getLastSyncHash();
 
-            if (lastHash == null || hash !== lastHash) {
+            if (lastHash == null || hash !== lastHash || hashLegacy !== lastHash) {
```
cscharf, line 91: I believe we need to combine the hash compares into a single boolean condition vs. multiple, otherwise one will likely always be true and lead to the re-sync I believe (my boolean brain is fighting with my reptilian brain atm).

```suggestion
            if (lastHash == null || (hash !== lastHash && hashLegacy !== lastHash)) {
```

cscharf (commented):
Awesome, just one last question/potential change.

src/services/sync.service.ts
MGibson1: Thanks, not enough sleep last night. I ended up writing a test to confirm behavior and you were right.

cscharf (approved):

PR #121, opened 2021-05-18
Use organization api key for auth

MGibson1:
# Overview

Related to bitwarden/jslib#382.

Transitions directory-connector over to using Organization Api Keys rather than a particular user's credentials.

> **NOTE:** This is a breaking change in all existing directory-connector configurations. Transitions should be to logout all existing clients

# Files Changed
* **apiKey.component/app.routing.module/app.module**: Pruned all the non-organization api key authentication stuff.
* **app.component**: no main thread log in work necessary with api keys
* **auth.guard/launch.guard/services.module**: use ApiKeyService as backing authed service
* **settings.component**: Remove organization selection -- we sync to the org whose api key we're authed with
* **bwdc.ts**: use ApiKeyService as backing authed service, also do not parallelize storageService writes.
* **program.ts**: Only allow api key logins and force option set.
* **auth.service**: This is an `AuthService` that overrides the api-style login to accept non-user api keys. It makes fewer assumptions about the returned authentication token.
* **sync.service**: Use the public api organization sync endpoint.

# Screenshots

### Login screen
image

# Testing considerations
A full regression suite of directory connector is necessary.

Emphasis should be placed on authenticating with valid and invalid api keys, upgrade paths to ensure we never try to sync with old credentials (expected behavior is to be logged out), and long-term syncs


cscharf:
@MGibson1 , from the screen shot, should we change the language on the login screen from referencing an `admin user` to something around the org API key instead?

src/app/accounts/apiKey.component.html
```diff
@@ -8,26 +8,23 @@
                     <h5 class="card-header">{{'logIn' | i18n}}</h5>
                     <div class="card-body">
                         <div class="form-group">
-                            <label for="email">{{'emailAddress' | i18n}}</label>
-                            <input id="email" type="text" name="Email" [(ngModel)]="email" class="form-control">
+                            <label for="client_id">Client ID</label>
```
cscharf, line 11: This needs to be localized.

src/app/accounts/apiKey.component.html
```diff
@@ -8,26 +8,23 @@
...
+                            <input id="client_id" type="password" name="Client Id" [(ngModel)]="clientId"
```
cscharf, line 12: generally HTML elements would have `name="..."` set without spaces, i.e. `name="ClientId"`.

src/app/accounts/apiKey.component.html
```diff
@@ -8,26 +8,23 @@
...
+                                class="form-control">
                         </div>
                         <div class="form-group">
                             <div class="row-main">
-                                <label for="masterPassword">{{'masterPass' | i18n}}</label>
-                                <input id="masterPassword" type="password" name="MasterPassword"
-                                    [(ngModel)]="masterPassword" class="form-control">
+                                <label for="client_secret">Client Secret</label>
```
cscharf, line 17: Same same.

src/app/accounts/apiKey.component.html
```diff
@@ -8,26 +8,23 @@
...
                             <div>
                                 <button type="submit" class="btn btn-primary" [disabled]="form.loading">
                                     <i class="fa fa-spinner fa-fw fa-spin" [hidden]="!form.loading"></i>
                                     <i class="fa fa-sign-in fa-fw" [hidden]="form.loading"></i>
-                                    {{'logIn' | i18n}}
-                                </button>
-                                <button type="button" class="btn btn-secondary ml-1" (click)="sso()">
-                                    <i class="fa fa-bank" aria-hidden="true"></i>
-                                    {{'enterpriseSingleSignOn' | i18n}}
+                                    {{'save' | i18n}}
```
cscharf, line 27: To try and understand the flow, the user could no longer log in "live" but only save credentials from the login screen? Would that change the behavior on usage (no longer use a GUI but only CLI)?

cscharf (commented):
Still working through the review, but wanted to drop a little bit of feedback sooner vs. later.

src/app/accounts/apiKey.component.html
MGibson1: I suppose you're still logging in, I just internalized it differently because you're saving your api key for later use.

GUI and CLI are both still fully functional. The only different in the GUI is you no longer choose an organization you are a part of to import to -- that's set by the api key you log in/save

MGibson1:
updated log in screen:
image

src/app/accounts/apiKey.component.html
MGibson1: GUI and CLI both fully available.

The only usage change is you can no longer select the organization to use from a drop down menu in GUI, because that's set by the clientId of your api key.

MGibson1:
force pushed to fix jslib conflict

cscharf (commented):
Looks good, just a few more comments/questions.

src/app/services/services.module.ts
```diff
@@ -53,6 +54,8 @@ import { StorageService as StorageServiceAbstraction } from 'jslib/abstractions/
 import { TokenService as TokenServiceAbstraction } from 'jslib/abstractions/token.service';
 import { UserService as UserServiceAbstraction } from 'jslib/abstractions/user.service';
 
+import { AuthService } from '../..//services/auth.service';
```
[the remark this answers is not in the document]
MGibson1: nope, thanks

src/program.ts
```diff
@@ -86,33 +90,14 @@ export class Program extends BaseProgram {
...
-                writeLn('    bwdc login john@example.com myPassword321');
-                writeLn('    bwdc login john@example.com myPassword321 --method 1 --code 249213');
-                writeLn('    bwdc login john@example.com --passwordfile passwd.txt --method 1 --code 249213');
-                writeLn('    bwdc login john@example.com --passwordenv MY_PASSWD --method 1 --code 249213');
-                writeLn('    bwdc login --sso');
-                writeLn('', true);
-            })
+            .command('login')
+            .description('Log into an organization account.')
             .action(async (email: string, password: string, options: program.OptionValues) => {
```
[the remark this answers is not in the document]
MGibson1: addressed. I don't like the direct parameter login, but it's probably worth keeping around

src/program.ts
```diff
@@ -86,33 +90,14 @@ export class Program extends BaseProgram {
...
-                writeLn('');
-                writeLn('    bwdc login');
-                writeLn('    bwdc login john@example.com myPassword321');
-                writeLn('    bwdc login john@example.com myPassword321 --method 1 --code 249213');
-                writeLn('    bwdc login john@example.com --passwordfile passwd.txt --method 1 --code 249213');
-                writeLn('    bwdc login john@example.com --passwordenv MY_PASSWD --method 1 --code 249213');
-                writeLn('    bwdc login --sso');
-                writeLn('', true);
-            })
+            .command('login')
```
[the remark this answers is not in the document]
MGibson1: you could still use env variables, but It's probably worth keeping the optional parameters as well.

cscharf (approved):
Looks good, thanks!

PR #125, opened 2021-06-01
Update Administrative Units API Endpoint

michaelklapper:
https://docs.microsoft.com/en-us/graph/api/administrativeunit-list-members?view=graph-rest-1.0#list-member-objects


MGibson1 (approved):
This is working on my end. It looks like MS release their API v1 without us updating.

Thanks for the help

PR #127, opened 2021-06-03
Add jslib as a "real" dependency

Hinton:
## Objective
Add jslib as "real" dependency which provides an easier way to sync dependency versions across our projects. 

- Split jslib into multiple projects.
- Change from using `postinstall` to `preinstall` since we expect the jslib packages.json files.
- Change all import paths from `jslib` to `jslib-common` or `jslib-angular`.

related to https://github.com/bitwarden/jslib/pull/363


src/bwdc.ts
```diff
@@ -11,23 +11,23 @@ import { KeytarSecureStorageService } from './services/keytarSecureStorage.servi
...
+import { ConsoleLogService } from 'jslib-node/cli/services/consoleLog.service';
+
+import { ApiKeyService } from 'jslib-common/services/apiKey.service';
+import { AppIdService } from 'jslib-common/services/appId.service';
+import { ConstantsService } from 'jslib-common/services/constants.service';
+import { ContainerService } from 'jslib-common/services/container.service';
+import { CryptoService } from 'jslib-common/services/crypto.service';
+import { EnvironmentService } from 'jslib-common/services/environment.service';
+import { NodeApiService } from 'jslib-node/services/nodeApi.service';
+import { NodeCryptoFunctionService } from 'jslib-common/services/nodeCryptoFunction.service';
```
MGibson1, line 24: Should this not be in jslib-node?

tsconfig.json
```diff
@@ -15,52 +15,24 @@
       "tldjs": [
         "jslib/src/misc/tldjs.noop"
       ],
-      "jslib/*": [
-        "jslib/src/*"
+      "jslib-common/*": [
+        "jslib/common/src/*"
       ],
-      "@angular/*": [
```
MGibson1, line 21: This and electron must come through the jslib dependencies, right?

tsconfig.json
```diff
@@ -15,52 +15,24 @@
...
-    "jslib/src/angular/directives/select-copy.directive.ts",
-    "jslib/src/importers",
-    "dist",
-    "dist-cli",
-    "jslib/dist",
-    "build",
-    "build-cli",
-    "jslib/spec"
+  "include": [
+    "src"
```
MGibson1, line 36: I believe you need `src-cli` in here too

webpack.main.js
```diff
@@ -23,7 +23,8 @@ const common = {
     resolve: {
         extensions: ['.tsx', '.ts', '.js'],
         alias: {
-            jslib: path.join(__dirname, 'jslib/src'),
+            "jslib-common": path.join(__dirname, 'jslib/common/src'),
+            "jslib-electron": path.join(__dirname, 'jslib/electron/src'),
```
MGibson1, line 27: is jslib-angular not required in webpack.renderer?

MGibson1 (commented):
Just a few questions

src/bwdc.ts
Hinton: Yes it should ...

tsconfig.json
Hinton: I don't know why we specified `@angular` manually here before since it should have never been required.

webpack.main.js
Hinton: That's the main.

Hinton: Switched to using `tsconfig-paths-webpack-plugin` as we do in the other projects.

MGibson1 (changes requested):
It looks like `prebuild-install` is now a required dev-dependency for CLI builds. We must not be pulling it in anymore because of the jslib dep changes. If you know better what's going on without adding the dependency that's fine too ;)

See `Package CLI` section of CLI build:
``` bash
'D:\a\directory-connector\directory-connector\node_modules\pkg\node_modules\.bin\prebuild-install' is not recognized as an internal or external command,
operable program or batch file.
'D:\a\directory-connector\directory-connector\node_modules\pkg\node_modules\.bin\prebuild-install' is not recognized as an internal or external command,
operable program or batch file.
'D:\a\directory-connector\directory-connector\node_modules\pkg\node_modules\.bin\prebuild-install' is not recognized as an internal or external command,
operable program or batch file.
```

Hinton:
@MGibson1 Seems `pkg` expects the `prebuild-install` to be installed in the `node_modules` of `pkg` which was the old behaviour in `npm` 6. I installed an older version of `prebuild-install` which seems to move the more recent version to `pkg`. 😢

PR #132, opened 2021-06-09
Improve okta group performance

MGibson1:
# Overview

Filter excluded/included groups as early as possible to avoid using up API calls and long waits.

We got some complaints concerning Okta-sync speed. Luckily the users have filters set, so we can improve things by filtering out groups earlier.

I also refactored the wait some to apply to `buildGroup` method. This saves us one wait, not a huge deal, but every bit helps 😃


cscharf (approved):
👍🏻

PR #111, opened 2021-04-21
Update ldapjs dependency

eliykat:
## Objective

In March 2019, we forked `ldapjs` in order to fix support for node 10 ([here](https://github.com/bitwarden/directory-connector/issues/13)). `ldapjs` added support for node 10 in August 2019 ([here](https://github.com/ldapjs/node-ldapjs/issues/524)). We should revert to using the official package and bump it to its latest version.

I've done the following for basic testing:
* reviewed release notes for any breaking changes ([here](https://github.com/ldapjs/node-ldapjs/releases)) - although these are very brief
* built the project and performed a test sync with a Jumpcloud instance

## Code changes

I specified the current version of `ldapjs` in `package.json` without using `^` or `~` modifiers. I'm not sure if we have any recommended practice around this.

I also updated `@types/ldapjs` to the current version.

This dependency does not appear in jslib or any other repo.

PR #133, opened 2021-06-15
Revert ldapjs update, use forked repo instead

eliykat:
## Objective

Revert the update to the `ldapjs` dependency which I did [here](https://github.com/bitwarden/directory-connector/pull/111). The update appears to have broken syncing for some users. Reverting to our fork until we can diagnose the issue.

This will need to be cherry-picked to `rc` after merge.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "bitwarden-directory-connector",
       "version": "0.0.0",
       "hasInstallScript": true,
       "license": "GPL-3.0",
@@ -23,7 +24,7 @@
         "form-data": "^4.0.0",
         "googleapis": "^73.0.0",
         "inquirer": "8.0.0",
-        "ldapjs": "2.2.4",
+        "ldapjs": "git+https://git@github.com/kspearrin/node-ldapjs.git",
         "lunr": "^2.3.9",
         "open": "^8.0.6",
         "proper-lockfile": "^4.1.2"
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -9102,34 +9128,51 @@
...
+      "integrity": "sha1-8rhCvguG2jNSeYUFsx68rlkNd9A=",
       "dependencies": {
-        "assert-plus": "^1.0.0"
+        "assert-plus": "0.1.5"
       },
       "engines": {
         "node": ">=0.8"
       }
     },
+    "node_modules/ldap-filter/node_modules/assert-plus": {
```
cscharf, line 9141: @eliykat , it would seem while you can revert the change to `package.json`, there have been SOOOO many other updates that it would be better to only revert that file and instead `npm i` to regen only the necessary portions of `package-lock.json` so as not to re-introduce prior packages or those handled by the `jslib` refactor since removed from here.

cscharf (changes requested):

package-lock.json
eliykat: Maybe I should have said "undo" rather than "revert"... I didn't use `git revert`, instead I did what you've suggested: I updated the line in `package.json` then ran `npm i` again, then committed changes.

I did a rough check of the changes in `package-lock.json` just now to make sure, and these changes all seem to be dependencies of `ldapjs`, I assume triggered by the downgrade.

cscharf (approved):

PR #134, opened 2021-06-16
Update okta-directory.service.ts

sburns90:
Replace part of the URL for the OKTA API call.
Specifically replaced 'users?filter=' with 'users?search=' and  'groups?filter=' with 'groups?search='
This enables the full utilization of the OKTA API to search for users and groups.

### DIFF
--- a/src/services/okta-directory.service.ts
+++ b/src/services/okta-directory.service.ts
@@ -68,7 +68,7 @@ export class OktaDirectoryService extends BaseDirectoryService implements IDirec
         const setFilter = this.createCustomSet(this.syncConfig.userFilter);
 
         this.logService.info('Querying users.');
-        const usersPromise = this.apiGetMany('users?filter=' + this.encodeUrlParameter(oktaFilter))
+        const usersPromise = this.apiGetMany('users?search=' + this.encodeUrlParameter(oktaFilter))
             .then((users: any[]) => {
                 for (const user of users) {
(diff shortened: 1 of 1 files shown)


cscharf (commented):
Hi @sburns90 , thank you for the contribution/PR! On quick review I do have some concerns in regards to the proposed changes.

In Okta, when filtering users or groups, the `filter` operation, while a little more narrow in scope (status, lastUpdated, id, profile.login, profile.email, profile.firstName, and profile.lastName), filters against the most up-to-date data and is recommended for critical flows. 

The `search` operation, on the other hand, has a wider scope for search but relies on a search index which Okta notes:

> Listing users with search should not be used as a part of any critical flows, such as authentication, to prevent potential data loss. Search results may not reflect the latest information, as this endpoint uses a search index which may not be up-to-date with recent updates to the object.

This may cause unexpected behavior for our current users syncing against Okta using directory connector and could be considered a breaking change or unintended behavior if delta syncs are in place and search index updates are sporadic.

I would prefer to see a new configuration option added for Okta to specify the use of `search` vs. `filter` with `filter` remaining the default. This would take a bit more work but would help reduce impact of a change like this for our existing customers using Okta through directory connector sync.

Hinton:
Hi @sburns90, it appears we've not heard from you about this issue in some time. While we are now closing this PR, please feel free to re-open a new PR and provide any further information that was requested.

PR #135, opened 2021-06-21
Refresh token with api key

MGibson1:
# Overview

Linked to bitwarden/jslib#414

bitwarden/jslib#382 and bitwarden/directory-connector#121 switched authentication to organization API key flow. However, this authentication does not include refresh tokens.

This work extends `ApiService` to override the token refresh flow with the api key authentication flow.

# Files changed
* **services.module.ts/bwdc.ts**: Update services to new `ApiService`. `refreshToken` is a little wonky. A method is needed due to a circular dependency on `apiKeyService` and `authService` to perform the refresh. I'm exporting the method because I wanted a single code source, but `ApiService` is already split higher upstream, requiring both `ApiService` and `NodeApiService` to be extended. The Compromise is to define the method in one of the `api.service.ts` files and export it while supplying the appropriate service dependencies.
* **api.service.ts**: New electron api service, overriding `doRefreshToken`
* **auth.service.ts**: Clear out client credentials when logging out.
* **nodeApi.service.ts**: New CLI apie service, overriding `doRefreshToken`

### DIFF
--- a/src/services/api.service.ts
+++ b/src/services/api.service.ts
@@ -2,6 +2,7 @@ import { ApiKeyService } from 'jslib-common/abstractions/apiKey.service';
 import { AuthService } from 'jslib-common/abstractions/auth.service';
 import { PlatformUtilsService } from 'jslib-common/abstractions/platformUtils.service';
 import { TokenService } from 'jslib-common/abstractions/token.service';
+
 import { ApiService as ApiServiceBase } from 'jslib-common/services/api.service';
 
 export async function refreshToken(apiKeyService: ApiKeyService, authService: AuthService) {
--- a/src/services/api.service.ts
+++ b/src/services/api.service.ts
@@ -1,0 +1,29 @@
+import { ApiKeyService } from 'jslib-common/abstractions/apiKey.service';
+import { AuthService } from 'jslib-common/abstractions/auth.service';
+import { PlatformUtilsService } from 'jslib-common/abstractions/platformUtils.service';
+import { TokenService } from 'jslib-common/abstractions/token.service';
+import { ApiService as ApiServiceBase } from 'jslib-common/services/api.service';
+
+export async function refreshToken(apiKeyService: ApiKeyService, authService: AuthService) {
+    try {
+        const clientId = await apiKeyService.getClientId();
+        const clientSecret = await apiKeyService.getClientSecret();
+        if (clientId != null && clientSecret != null) {
+            await authService.logInApiKey(clientId, clientSecret);
+        }
+    } catch (e) {
+        return Promise.reject(e);
+    }
+}
+
+export class ApiService extends ApiServiceBase {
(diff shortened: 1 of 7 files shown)


src/services/api.service.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export class ApiService extends ApiServiceBase {
+    constructor(tokenService: TokenService, platformUtilsService: PlatformUtilsService,
+        private refreshTokenCallback: () => Promise<void>, logoutCallback: (expired: boolean) => Promise<void>,
+        customUserAgent: string = null) {
+        super(tokenService, platformUtilsService, logoutCallback, customUserAgent);
+    }
+
+    doRefreshToken(): Promise<void> {
+        return this.refreshTokenCallback();
```
Hinton, line 27: Not a fan of the callback, but we probably don't want to tackle this now. Feels like we should extract the `doRefreshToken` to another service.

src/services/api.service.ts
```diff
@@ -0,0 +1,29 @@
+import { ApiKeyService } from 'jslib-common/abstractions/apiKey.service';
+import { AuthService } from 'jslib-common/abstractions/auth.service';
+import { PlatformUtilsService } from 'jslib-common/abstractions/platformUtils.service';
+import { TokenService } from 'jslib-common/abstractions/token.service';
+import { ApiService as ApiServiceBase } from 'jslib-common/services/api.service';
```
cscharf, line 5: Maybe just separating the services from the abstractions into a new block for imports would be good, not important but would follow convention.

src/services/nodeApi.service.ts
```diff
@@ -0,0 +1,14 @@
+import { PlatformUtilsService } from 'jslib-common/abstractions/platformUtils.service';
+import { TokenService } from 'jslib-common/abstractions/token.service';
+import { NodeApiService as NodeApiServiceBase } from 'jslib-node/services/nodeApi.service';
```
cscharf, line 3: Same with the whole block thing.

src/services/nodeApi.service.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export class NodeApiService extends NodeApiServiceBase {
+    constructor(tokenService: TokenService, platformUtilsService: PlatformUtilsService,
+        private refreshTokenCallback: () => Promise<void>, logoutCallback: (expired: boolean) => Promise<void>, customUserAgent: string = null) {
```
cscharf, line 7: Formatting nit... wrap the last argument to a new line as it exceeds the line length a bit and follows the two-per-line constructor argument pattern elsewhere.

cscharf (commented):
This looks good, other than the few formatting nits.

MGibson1:
Sorry, forgot to tag this, but this will need cherry-picking to `rc`

cscharf (approved):

MGibson1:
cherry picked to `rc`

PR #136, opened 2021-06-24
Error on duplicate emails

MGibson1:
# Overview

Fixes https://app.asana.com/0/1199950673439194/1200501375491548/f. Some LDAP configurations allow for users to have duplicated email addresses. Bitwarden does not allow this and errors.

This work early detects this failure and displays an actionable error message.

# Files Changed
* **package.json**: Updates `npm` scripts to include source maps to main on dev builds. This is the same work done in bitwarden/desktop#946
* **webpack.main.js**: include source maps for main process in dev builds. This is the same work done in bitwarden/desktop#946
* **messages.json**: Message displayed explaining duplicate email failure.
* **sync.service.ts**: Determine duplicated emails and throw if any are found


cscharf:
@MGibson1 , seems like one of those updates is causing `keytar` to fail the CLI build. Not sure if it's related.

MGibson1:
I'm not sure what that `keytar` failure was, but rerunning got us back to the more typical CLI failure of windows not binding the command.

I'll make an issue to fix this action in the pipeline so we can get our check marks back

src/services/sync.service.ts
```diff
@@ -55,6 +55,11 @@ export class SyncService {
                 this.flattenUsersToGroups(groups, groups);
             }
 
+            const duplicateEmails = this.findDuplicateUserEmails(users);
+            if (duplicateEmails.length > 0) {
+                throw new Error(this.i18nService.t('duplicateEmails') + '\n' + duplicateEmails.join('\n'));
```
cscharf, line 60: If this is going to show up in the little red error toast, on a large sync this could be 1 or 1k's of email addresses, probably ideal to cut this off @ only the first _n_ and then use something like `+ '... (+${duplicateEmails.length - n} more)'` or something like that

cscharf: I like having the emails in the error message, it's super helpful, and at risk of overcomplicating the creation of an error message, but at the same time if it's more than say, 3, then I think it's worth it to truncate the message and append a count kind of thing.

cscharf (commented):
Just some general feedback.... Also, perhaps worthwhile to split out the package + webpack changes (which is awesome they'll match desktop) into a separate PR?

src/services/sync.service.ts
MGibson1: yeah, you're probably right. That's the kind of polish we should strive for

MGibson1:
I removed the sourcemap changes, you're right, those shouldn't be picked to `rc`, which this will be.

cscharf (approved):

MGibson1:
cherry picked to `rc`

PR #144, opened 2021-08-18
Fix Azure client errors not displaying properly

eliykat:
## Objective

Fix this behaviour, as requested by the good folks over in customer success:

>When configuring BWDC to use Azure Active Directory sync, and an invalid Application ID or Secret Key for the Azure AD auth is provided, BWDC displays an "unknown error has occurred" error card and does not log any network requests or errors on the console. BWDC should display "invalid AppID or secret key" instead.

## Code changes

In `azureDirectoryService.init()`, we define a callback which is responsible for authenticating and getting the access token. After its work is completed, it calls the `done` function with either an error or the access token.

It looked like we were doing this already, but we were getting stack traces rather than nice error messages, and that was changed to a generic error message by `appApiAction`. It turns out that `done` should receive an `Error` object, not just a string containing the error message. After changing error strings to `Error` objects, they display as expected.

Azure provides fairly verbose messages, so for better UX I have only displayed the first line in the toast, and printed the rest to the console. This code only applies to Azure authentication errors so I felt I could be reasonably specific.

Here's an example of how it looks:

Screen Shot 2021-08-18 at 4 28 45 pm

For comparison, here's how it looks if we don't do this (keeping in mind you have about 3 seconds to parse this before it disappears):

Screen Shot 2021-08-18 at 4 05 35 pm

### DIFF
--- a/src/services/azure-directory.service.ts
+++ b/src/services/azure-directory.service.ts
@@ -386,7 +386,7 @@ export class AzureDirectoryService extends BaseDirectoryService implements IDire
             authProvider: done => {
                 if (this.dirConfig.applicationId == null || this.dirConfig.key == null ||
                     this.dirConfig.tenant == null) {
-                    done(this.i18nService.t('dirConfigIncomplete'), null);
+                    done(new Error(this.i18nService.t('dirConfigIncomplete')), null);
                     return;
                 }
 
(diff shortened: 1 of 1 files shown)


cscharf (approved):
Nice! Looks good and good decision on the first line display.

PR #152, opened 2021-09-09
Ignore duplicate users with same externalId

eliykat:
## Objective

We made some changes [here](https://github.com/bitwarden/directory-connector/pull/136) to throw an error if the directory service contains 2 or more users with the same email address. This is because we don't support duplicate email addresses in Bitwarden.

Some users have reported false positives, where the error was being thrown even though they didn't think that they had more than 1 user with the same email address.

I was able to reproduce this by simply disabling a user in Azure AD. This caused the disabled user to be returned by both the `getCurrentUsers` and the `getDeletedUsers` methods, which then caused the duplicate entry.

## Code changes

* `sync.service` - only throw an error if multiple `userEntry`s have the same email address but _different_ `externalId`s. `externalId`s should be unique to each user, so if the email address and `externalId` are identical, it's probably the same user and we can safely remove the duplicates without throwing the error. This should reduce the incidence of false positives.
* `azure-directory.service` - change the boolean logic of `getDeletedUsers` so that disabled (but not deleted) users are excluded from this user set. This should remove the source of the duplicate to begin with. I wasn't 100% sure which set disabled users should go in, because they are neither "current" nor "deleted", but (a) I don't think it matters too much because both sets are concatenated at the end, so as long as there's no overlap, and (b) at a glance, I _think_ our LDAP and GSuite services work this way, so this seemed the most consistent.

### DIFF
--- a/src/services/azure-directory.service.ts
+++ b/src/services/azure-directory.service.ts
@@ -144,7 +144,7 @@ export class AzureDirectoryService extends BaseDirectoryService implements IDire
                         continue;
                     }
                     const entry = this.buildUser(user);
-                    if (!entry.disabled && !entry.deleted) {
+                    if (!entry.deleted) {
                         continue;
                     }
                     if (await this.filterOutUserResult(setFilter, entry, false)) {
--- a/src/services/sync.service.ts
+++ b/src/services/sync.service.ts
@@ -117,16 +117,16 @@ export class SyncService {
         const processedUsers = new Map<string, string>();
         const duplicateEmails = new Array<string>();
 
-        // UserEntrys with the same email and externalId are removed but otherwise ignored
-        // UserEntrys with the same email but different externalIds will throw an error
+        // UserEntrys with the same email are ignored if their properties are the same
+        // UserEntrys with the same email but different properties will throw an error
         users.forEach(u => {
             if (processedUsers.has(u.email)) {
-                if (u.externalId == null || processedUsers.get(u.email) != u.externalId) {
+                if (processedUsers.get(u.email) != JSON.stringify(u)) {
                     duplicateEmails.push(u.email);
                 }
             } else {
                 uniqueUsers.push(u);
-                processedUsers.set(u.email, u.externalId);
+                processedUsers.set(u.email, JSON.stringify(u));
             }
         });
 
--- a/src/services/sync.service.ts
+++ b/src/services/sync.service.ts
@@ -121,7 +121,7 @@ export class SyncService {
         // UserEntrys with the same email but different externalIds will throw an error
         users.forEach(u => {
             if (processedUsers.has(u.email)) {
-                if (processedUsers.get(u.email) != u.externalId) {
+                if (processedUsers.get(u.email) != u.externalId || u.externalId == null) {
                     duplicateEmails.push(u.email);
                 }
             } else {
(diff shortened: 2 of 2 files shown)


src/services/azure-directory.service.ts
```diff
@@ -144,7 +144,7 @@ export class AzureDirectoryService extends BaseDirectoryService implements IDire
                         continue;
                     }
                     const entry = this.buildUser(user);
-                    if (!entry.disabled && !entry.deleted) {
+                    if (!entry.deleted) {
```
MGibson1, line 147: The difference is do we want to only grab disabled users since the last call, unless force is on. I think probably you're right in your decision to always disable

src/services/sync.service.ts
```diff
@@ -118,17 +112,32 @@ export class SyncService {
...
-        const duplicatedEmails = new Array<string>();
-        users.reduce((agg, user) => {
-            if (agg.includes(user.email) && !duplicatedEmails.includes(user.email)) {
-                duplicatedEmails.push(user.email);
+    private removeDuplicateUsers(users: UserEntry[]) {
+        const uniqueUsers = new Array<UserEntry>();
+        const processedUsers = new Map<string, string>();
+        const duplicateEmails = new Array<string>();
+
+        // UserEntrys with the same email and externalId are removed but otherwise ignored
```
MGibson1, line 120: I'm afraid of removing these entries with just this requirement.

These Entries are glorified structs: Can we instead serialize them and compare the serialized strings? That way we remove only if they really are equivalent.

If they're not, something weird happened and we can't determine which is valid and which is not.

MGibson1 (changes requested):
I know we talked about this a bit, but seeing it written out got me more worried about removing entries. Comment discusses new idea

src/services/sync.service.ts
eliykat: Yep good call, better to be careful before discarding entries. I've changed this to use `JSON.stringify` and it works as expected. I checked the following caveats but I think they're OK:

* doesn't compare methods or stuff up the prototype chain - fine with us
* the order that properties are stringified can be affected by the order they're initialized/assigned values. However, all the `UserEntry`s are initialized in a consistent manner at the same time, so this should be fine in our case.

src/services/azure-directory.service.ts
eliykat: Not sure I follow you - this `if` statement will exclude a `UserEntry` from the `DeletedUsers` set if the condition is true.

At the moment, disabled users are included in the set because for them, `entry.disabled === true && entry.deleted === false`, so this condition is false. And then they're included again in the `CurrentUsers` set because they're not deleted. So the idea is to take them out of the `DeletedUsers` set to avoid the duplicate.

Let me know if I've misunderstood you.

MGibson1: I meant to explain what the difference was between `getUsers` and `getDeletedUsers`. Deleted users are pulled using a change token from the last sync (unless force is on), which allows MS to only tell us about updates. `getUsers` just grabs all users.

Now that I re-read that comment, it makes no sense in a vacuum. I was responding to my inner monologue, I guess. What's even worse is I don't think this comment adds anything since I'm just agreeing with you -- sorry

MGibson1 (approved):
I'm too used to c#, I wasn't thinking about the order of properties -- good catch.

src/services/azure-directory.service.ts
eliykat: I hadn't considered that, so good to know! And don't worry, we spend all day in our own heads 😆

PR #159, opened 2021-09-27
Update pipeline to new model

vgrassia:
This PR is to update our `build` and `release` pipelines to our new workflow model.


.github/workflows/release.yml
```diff
@@ -29,118 +22,59 @@ jobs:
...
+          fi
+        shell: bash
 
-      - name: Download latest RC build artifacts
+      - name: Download all artifacts
         uses: dawidd6/action-download-artifact@b9571484721e8187f1fd08147b497129f8972c74  # v2.14.0
         with:
           workflow: build.yml
           workflow_conclusion: success
           branch: rc
```
joseph-flinn, line 54: Do we need all of the artifacts? Or can we download only specific ones with our action: https://github.com/bitwarden/gh-actions/tree/master/download-artifacts

joseph-flinn (commented):
I'm wondering if we can further minimize the building in the release by using pre-packaged release assets for the Windows-GUI, linux, and Mac. the only difference between `npm run dist:lin` and `npm run publish:lin` is that publish uploads the asset to the latest release instead of us doing it "manually"

.github/workflows/release.yml
vgrassia: I think all artifacts should be uploaded as that's what we're currently doing.  I also removed the jobs in the `release` workflow.  I didn't know the difference between `publish` and `dist` and now I see they are doing the same thing except for the upload.

joseph-flinn (approved):

PR #174, opened 2021-11-08
Update electron to 14.2.0

djsmith85:
## Type of change
- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
Bump jslib and update to electron 14.2.0

## Code changes
* **jslib:** Pull in newest changes to jslib
* **package-lock.json:** Changes after running `npm i`
* **services/auth.service.ts:** Added missing dependency param `CryptoFunctionService` for ctor of `AuthServiceBase`
* **bwdc.ts:** Added `CryptoFunctionService` to ctor of `AuthService`
* **app/services/services.module.ts:** Added `CryptoFunctionService` to ctor of `AuthService`

## Testing requirements
Usual regression testing

## Before you submit
- [X] I have checked for **linting** errors (`npm run lint`) (required)
   - Still two issues that actually should be ignored console.error in `azure-directory.service.ts` and a equality-check in `sync.services.ts`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -131,7 +131,7 @@
       "dependencies": {
         "@bitwarden/jslib-common": "file:../common",
         "@nodert-win10-rs4/windows.security.credentials.ui": "^0.4.4",
-        "electron": "14.0.1",
+        "electron": "14.2.0",
         "electron-log": "4.4.1",
         "electron-store": "8.0.1",
         "electron-updater": "4.3.9",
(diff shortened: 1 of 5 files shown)


eliykat (changes requested):
Can you fix the tslint errors while you're here? Even if they should be ignored, we should make sure tslint knows that.

```
src/services/azure-directory.service.ts:427:29
ERROR: 427:29  no-console     Calls to 'console.error' are not allowed.
```

This already has an ignore rule above it, but there's a space after the colon which is preventing it from working. Let's fix it up while we're here.

```
src/services/sync.service.ts:124:49
ERROR: 124:49  triple-equals  != should be !==
```

This was me 🙈  This is just a string comparison, I think the tslint error is correct. While I'm 99% sure it won't cause regressions to change it, I should probably do it in a separate PR so it's not buried. We can leave this for now.

djsmith85:
@eliykat  Fixed the tslint ignore rule causing the linter to complain about the console.error()

I also wasn't sure if the single equality check in sync.service.ts was intended, so to not cause regressions I left it as is. We'll get to it in another PR.

eliykat (approved):

PR #181, opened 2021-11-29
Don't flag deleted users as duplicates

differsthecat:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fixes an issue where a duplicate email error is being thrown. After a lot of investigation, the root cause of this has been determined to be deleted users; AD unfortunately allows emails to be duplicated for deleted users. Additional context here: https://app.asana.com/0/1169444489336079/1201410771133282/f

## Code changes

* **sync.service.ts** Add a condition to only add a user to the duplicateEmails array if the user is not deleted.

## Testing requirements

Please test around the syncing functionality and make sure that duplicates aren't flagged unexpectedly.
The change was done in a shared function in the sync service that all directory types use.

These are the expected outcomes when testing for duplicate emails:
- Two or more deleted users with the same email will not throw a duplicate email error.
- If an active user and a deleted user share the same email, a duplicate email error will be thrown.
- If two or more active users share an email, a duplicate email error will be thrown.

Feel free to reach out to me (Robyn) for some ways to set up the above conditions in Active Directory.

## Before you submit

### DIFF
--- a/src/services/sync.service.ts
+++ b/src/services/sync.service.ts
@@ -127,7 +127,7 @@ export class SyncService {
                 }
             } else {
                 if (!u.deleted) {
-                    // UserEntrys that are not deleted but there's a deleted entry with same email error
+                    // Check that active UserEntry does not conflict with a deleted UserEntry
                     if (processedDeletedUsers.has(u.email)) {
                         duplicateEmails.push(u.email);
                     } else {
--- a/src/services/sync.service.ts
+++ b/src/services/sync.service.ts
@@ -115,20 +115,30 @@ export class SyncService {
     private removeDuplicateUsers(users: UserEntry[]) {
         const uniqueUsers = new Array<UserEntry>();
         const processedUsers = new Map<string, string>();
+        const processedDeletedUsers = new Map<string, string>();
         const duplicateEmails = new Array<string>();
 
         // UserEntrys with the same email are ignored if their properties are the same
-        // UserEntrys with the same email but different properties will throw an error
+        // UserEntrys with the same email but different properties will throw an error 
         users.forEach(u => {
             if (processedUsers.has(u.email)) {
                 if (processedUsers.get(u.email) !== JSON.stringify(u)) {
-                    if (!u.deleted) {
-                        duplicateEmails.push(u.email);
-                    }
+                    duplicateEmails.push(u.email);
                 }
             } else {
-                uniqueUsers.push(u);
-                processedUsers.set(u.email, JSON.stringify(u));
+                if (!u.deleted) {
+                    // UserEntrys that are not deleted but there's a deleted entry with same email error
+                    if (processedDeletedUsers.has(u.email)) {
+                        duplicateEmails.push(u.email);
+                    } else {
+                        processedUsers.set(u.email, JSON.stringify(u));
+                        uniqueUsers.push(u);
+                    }
+                } else {
+                    // UserEntrys with the same email but are all deleted are ignored but tracked
+                    processedDeletedUsers.set(u.email, JSON.stringify(u));
+                    uniqueUsers.push(u);
+                }
             }
         });
 
--- a/src/services/sync.service.ts
+++ b/src/services/sync.service.ts
@@ -119,7 +119,7 @@ export class SyncService {
         const duplicateEmails = new Array<string>();
 
         // UserEntrys with the same email are ignored if their properties are the same
-        // UserEntrys with the same email but different properties will throw an error 
+        // UserEntrys with the same email but different properties will throw an error, unless they are all in a deleted state.
         users.forEach(u => {
             if (processedActiveUsers.has(u.email)) {
                 if (processedActiveUsers.get(u.email) !== JSON.stringify(u)) {
--- a/src/services/sync.service.ts
+++ b/src/services/sync.service.ts
@@ -122,7 +122,9 @@ export class SyncService {
         users.forEach(u => {
             if (processedUsers.has(u.email)) {
                 if (processedUsers.get(u.email) != JSON.stringify(u)) {
-                    duplicateEmails.push(u.email);
+                    if (!u.deleted) {
+                        duplicateEmails.push(u.email);
+                    }
                 }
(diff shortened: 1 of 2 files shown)


differsthecat:
@MGibson1 , @eliykat It looks like you have both worked on this part of the code before, I would appreciate your thoughts on this!

eliykat:
Is this a bug? Directory Connector deprovisions users as well. It seems like the usual concern about duplicate email addresses applies here as well. That is, emails are unique identifiers in Bitwarden, and Bitwarden can't tell two accounts apart if they use the same email.

For example:
* User A is active in the directory service and has been provisioned with a Bitwarden account
* User A is then deleted from the directory service, and User B is added using the same email
* bwdc tries to sync these changes - I'm not sure what would happen at the moment, but my guess is that no deprovisioning would occur. User A would still have access to their Bitwarden account because the email address is still present in AD. User B does not get invited because they don't appear as a new user. (Much worse than a duplicate email error.)

Note that the `removeDuplicateUsers` method compares a stringified version of the `UserEntry` object, so we're already comparing all user properties that are available to us.

I think this is either a problem with the user's directory service configuration, or a larger problem with bwdc. Not sure what the answer is.

differsthecat:
@eliykat Those are great points! With what is currently happening there is definitely a bug. We may have to approach it differently than I did here though.

There are a couple scenarios where the client ends up in a state where they can not sync anything at all if there are users that are deleted that have email conflicts with each other. 
Another way we can tackle this is to apply the users filter to the deleted users query. Currently all deleted users are grabbed, regardless of if they are deleted or not. So even if you filter down users to one single user, if there are conflicts in deleted users between themselves, it throws the duplicate error.

We can also change how we look for duplicates and throw an error if a non-deleted user conflicts with a deleted user, but allow deleted users to be duplicated between each other.

differsthecat:
@eliykat Here's the code in question that neglects to apply the filter to the deleted users:
https://github.com/bitwarden/directory-connector/blob/abc68e8ef93642490b4e0ccfbb8cee0ef0a44719/src/services/ldap-directory.service.ts#L82
Up on line 70 of this function, the supplied filter is applied to the regular users query.

differsthecat:
> @eliykat Here's the code in question that neglects to apply the filter to the deleted users:
> 
> https://github.com/bitwarden/directory-connector/blob/abc68e8ef93642490b4e0ccfbb8cee0ef0a44719/src/services/ldap-directory.service.ts#L82
> 
> 
> Up on line 70 of this function, the supplied filter is applied to the regular users query.

I spent some time playing with this, and an issue with adding the user's filter to the deleted users query is that in AD deleted users are missing many fields that non-deleted users have (such as sAMAccountName, givenName) so deleted users can get removed when they are not expected to. This would need a more creative solution and is out of scope of this for now, so I reverted those changes.

src/services/sync.service.ts
```diff
@@ -115,18 +115,30 @@ export class SyncService {
...
+                        duplicateEmails.push(u.email);
+                    } else {
+                        processedUsers.set(u.email, JSON.stringify(u));
+                        uniqueUsers.push(u);
+                    }
+                } else {
+                    // UserEntrys with the same email but are all deleted are ignored but tracked
+                    processedDeletedUsers.set(u.email, JSON.stringify(u));
+                    uniqueUsers.push(u);
+                }
```
differsthecat, line 141: This may be a bit tricky to read, but:
- If a user is active (not deleted) and there is another user with the same email (deleted or not), the duplicate error is thrown.
- If a user is deleted and there are any other users that are also deleted with the same email, no error is thrown.

This will fix the case that clients are seeing where deleted users that don't conflict with active users are causing a sync to fail.

src/services/sync.service.ts
```diff
@@ -115,18 +115,30 @@ export class SyncService {
     private removeDuplicateUsers(users: UserEntry[]) {
         const uniqueUsers = new Array<UserEntry>();
         const processedUsers = new Map<string, string>();
+        const processedDeletedUsers = new Map<string, string>();
         const duplicateEmails = new Array<string>();
 
         // UserEntrys with the same email are ignored if their properties are the same
-        // UserEntrys with the same email but different properties will throw an error
+        // UserEntrys with the same email but different properties will throw an error
```
eliykat, line 122: Update to describe new logic, e.g.

```suggestion
        // UserEntrys with the same email but different properties will throw an error, unless they are all in a deleted state.
```

src/services/sync.service.ts
```diff
@@ -114,19 +114,31 @@ export class SyncService {
...
+                if (!u.deleted) {
+                    // UserEntrys that are not deleted but there's a deleted entry with same email error
+                    if (processedDeletedUsers.has(u.email)) {
+                        duplicateEmails.push(u.email);
+                    } else {
+                        processedActiveUsers.set(u.email, JSON.stringify(u));
+                        uniqueUsers.push(u);
+                    }
+                } else {
+                    // UserEntrys with the same email but are all deleted are ignored but tracked
```
eliykat, line 138:
```suggestion
                    // UserEntrys with duplicate email will not throw an error if they are all deleted. They will be synced.
```

src/services/sync.service.ts
```diff
@@ -114,19 +114,31 @@ export class SyncService {
...
-                if (processedUsers.get(u.email) != JSON.stringify(u)) {
+            if (processedActiveUsers.has(u.email)) {
+                if (processedActiveUsers.get(u.email) !== JSON.stringify(u)) {
                     duplicateEmails.push(u.email);
                 }
             } else {
-                uniqueUsers.push(u);
-                processedUsers.set(u.email, JSON.stringify(u));
+                if (!u.deleted) {
+                    // UserEntrys that are not deleted but there's a deleted entry with same email error
```
eliykat, line 130:
```suggestion
                    // Check that active UserEntry does not conflict with a deleted UserEntry
```

eliykat (changes requested):

eliykat (approved):
Nice work!

PR #192, opened 2021-12-16
[refactor] Implement StateService

addisonbeck:
* Depends on https://github.com/bitwarden/jslib/pull/584

## Type of change
- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective
Implements the new state service model change made in jslib.
Depends on: https://github.com/bitwarden/jslib/pull/584

## Code changes
Most changes are swapping the old state service implementation to the new one, with the addition of a migrator to migrate the old state structure to the new one and an extendio of state service for directory connector specific state.

## Testing requirements
**IMPORTANT** Please make a backup of `data.json` before running. I tested this locally with the existing `data.json` file I had, but I may have missed some things.

## Before you submit


.github/workflows/build.yml
```diff
@@ -98,7 +98,11 @@ jobs:
 
       - name: Version Test
         run: |
-          sudo apt install libsecret-1-0
+          sudo apt install libsecret-1-0 dbus-x11 gnome-keyring
+          eval $(dbus-launch --sh-syntax)
+          
+          eval $(echo -n "" | /usr/bin/gnome-keyring-daemon --login)
+          eval $(/usr/bin/gnome-keyring-daemon --components=secrets --start)
```
joseph-flinn, line 105:
```suggestion
          dbus-launch --sh-syntax
          
          echo -n "" | /usr/bin/gnome-keyring-daemon --login
          /usr/bin/gnome-keyring-daemon --components=secrets --start
```

I don't believe `eval` is needed here. It is normally used to evaluate a command that is stored in a variable as a string like
```bash
var="ls -alth"
eval $var
```

differsthecat: I'll give it a try and see if it is still happy with it!

differsthecat: It's erroring again, I'm going to add the `eval`'s back

joseph-flinn: Weird. Ok, we can keep the `eval`'s then.

MGibson1: At the least `eval $(dbus-launch --sh-syntax)` is necessary. The application outputs a set of env variables about the currently running dbus session.

I don't believe the others are as there is no output.

src/program.ts
```diff
@@ -16,10 +16,10 @@ import { UpdateCommand } from "jslib-node/cli/commands/update.command";
 
 import { BaseProgram } from "jslib-node/cli/baseProgram";
 
-import { ApiKeyService } from "jslib-common/abstractions/apiKey.service";
 import { Response } from "jslib-node/cli/models/response";
 import { StringResponse } from "jslib-node/cli/models/response/stringResponse";
 
+import { HtmlStorageLocation } from "jslib-common/enums/htmlStorageLocation";
```
MGibson1, line 22: This enum scares me. DC uses electron and lowdb for storage, what's up with the name?

MGibson1: Also, it's unused and can be removed

src/services/configuration.service.ts
```diff
@@ -75,154 +58,150 @@ export class ConfigurationService {
...
   }
 
-  saveLastGroupSyncDate(date: Date) {
+  async saveLastGroupSyncDate(date: Date): Promise<void> {
     if (date == null) {
-      return this.storageService.remove(Keys.lastGroupSync);
+      await this.stateService.setLastGroupSync(null);
     } else {
-      return this.storageService.save(Keys.lastGroupSync, date);
+      await this.stateService.setLastGroupSync(date);
```
MGibson1, line 175: No purpose to this `if` statement anymore

src/services/configuration.service.ts
```diff
@@ -75,154 +58,150 @@ export class ConfigurationService {
...
   }
 
-  saveLastSyncHash(hash: string) {
+  async saveLastSyncHash(hash: string): Promise<void> {
     if (hash == null) {
-      return this.storageService.remove(Keys.lastSyncHash);
+      await this.stateService.setLastSyncHash(null);
     } else {
-      return this.storageService.save(Keys.lastSyncHash, hash);
+      await this.stateService.setLastSyncHash(hash);
```
MGibson1, line 187: No purpose to this `if` statement anymore

src/services/configuration.service.ts
```diff
@@ -75,154 +58,150 @@ export class ConfigurationService {
...
   }
 
-  saveLastUserSyncDate(date: Date) {
+  async saveLastUserSyncDate(date: Date): Promise<void> {
     if (date == null) {
-      return this.storageService.remove(Keys.lastUserSync);
+      return await this.stateService.setLastUserSync(null);
     } else {
-      return this.storageService.save(Keys.lastUserSync, date);
+      return await this.stateService.setLastUserSync(date);
```
MGibson1, line 163: No purpose to this `if` statement anymore

src/services/configuration.service.ts
```diff
@@ -75,154 +58,150 @@ export class ConfigurationService {
...
   }
 
-  saveGroupDeltaToken(token: string) {
+  async saveGroupDeltaToken(token: string): Promise<void> {
     if (token == null) {
-      return this.storageService.remove(Keys.groupDelta);
+      await this.stateService.setGroupDelta(null);
     } else {
-      return this.storageService.save(Keys.groupDelta, token);
+      await this.stateService.setGroupDelta(token);
```
MGibson1, line 151: No purpose to this `if` statement anymore

src/services/configuration.service.ts
```diff
@@ -75,154 +58,150 @@ export class ConfigurationService {
...
   }
 
-  saveUserDeltaToken(token: string) {
+  async saveUserDeltaToken(token: string): Promise<void> {
     if (token == null) {
-      return this.storageService.remove(Keys.userDelta);
+      return this.stateService.setUserDelta(null);
     } else {
-      return this.storageService.save(Keys.userDelta, token);
+      return this.stateService.setUserDelta(token);
```
MGibson1, line 139: No purpose to this `if` statement anymore

src/services/configuration.service.ts
```diff
@@ -75,154 +58,150 @@ export class ConfigurationService {
...
           }
+          await this.stateService.setGsuiteConfiguration(savedConfig);
           break;
         case DirectoryType.OneLogin:
           if (savedConfig.clientSecret == null) {
-            await this.secureStorageService.remove(Keys.oneLogin);
+            await this.stateService.setOneLoginKey(null);
           } else {
-            await this.secureStorageService.save(Keys.oneLogin, savedConfig.clientSecret);
+            await this.stateService.setOneLoginKey(savedConfig.clientSecret);
```
MGibson1, line 101: No purpose to this `if` statement anymore

src/services/configuration.service.ts
```diff
@@ -75,154 +58,150 @@ export class ConfigurationService {
...
           }
+          await this.stateService.setAzureConfiguration(savedConfig);
           break;
         case DirectoryType.Okta:
           if (savedConfig.token == null) {
-            await this.secureStorageService.remove(Keys.okta);
+            await this.stateService.setOktaKey(null);
           } else {
-            await this.secureStorageService.save(Keys.okta, savedConfig.token);
+            await this.stateService.setOktaKey(savedConfig.token);
```
MGibson1, line 81: No purpose to this `if` statement anymore

src/services/configuration.service.ts
```diff
@@ -75,154 +58,150 @@ export class ConfigurationService {
...
           }
+          await this.stateService.setLdapConfiguration(savedConfig);
           break;
         case DirectoryType.AzureActiveDirectory:
           if (savedConfig.key == null) {
-            await this.secureStorageService.remove(Keys.azure);
+            await this.stateService.setAzureKey(null);
           } else {
-            await this.secureStorageService.save(Keys.azure, savedConfig.key);
+            await this.stateService.setAzureKey(savedConfig.key);
```
MGibson1, line 72: No purpose to this `if` statement anymore

src/services/configuration.service.ts
```diff
@@ -75,154 +58,150 @@ export class ConfigurationService {
       switch (type) {
         case DirectoryType.Ldap:
           if (savedConfig.password == null) {
-            await this.secureStorageService.remove(Keys.ldap);
+            await this.stateService.setLdapKey(null);
           } else {
-            await this.secureStorageService.save(Keys.ldap, savedConfig.password);
+            await this.stateService.setLdapKey(savedConfig.password);
```
MGibson1, line 63: No purpose to this `if` statement anymore

src/services/configuration.service.ts
```diff
@@ -75,154 +58,150 @@ export class ConfigurationService {
...
     if (currentId !== id) {
       await this.clearStatefulSettings();
     }
 
     if (id == null) {
-      return this.storageService.remove(Keys.organizationId);
+      await this.stateService.setOrganizationId(null);
     } else {
-      return this.storageService.save(Keys.organizationId, id);
+      await this.stateService.setOrganizationId(id);
```
MGibson1, line 204: No purpose to this `if` statement anymore

src/services/auth.service.ts
```diff
@@ -88,9 +92,33 @@ export class AuthService extends AuthServiceBase {
...
     result.resetMasterPassword = tokenResponse.resetMasterPassword;
-    await this.tokenService.setToken(tokenResponse.accessToken);
-    await this.apiKeyService.setInformation(clientId, clientSecret);
-
+    await this.stateService.addAccount(
+      new Account({
+        profile: {
+          ...new AccountProfile(),
+          ...{
+            userId: clientId,
```
MGibson1, line 100: seems a little odd to use clienId here, we don't do anything expecting a Guid, right?

Maybe to be safe this should be `entityId` anyways?

src/services/state.service.ts
```diff
@@ -0,0 +1,434 @@
...
+    options = this.reconcileOptions(options, await this.defaultSecureStorageOptions());
+    if (options?.userId == null) {
+      return;
+    }
+    await this.secureStorageService.save(
+      `${options.userId}_${SecureStorageKeys.ldap}`,
+      value,
+      options
+    );
+  }
```
MGibson1, line 60: All this repeated `options =...` stuff is begging for a template method that takes a callback to `get` or `save`

src/services/state.service.ts
```diff
@@ -0,0 +1,434 @@
...
+
+  async setLdapKey(value: string, options?: StorageOptions): Promise<void> {
+    options = this.reconcileOptions(options, await this.defaultSecureStorageOptions());
+    if (options?.userId == null) {
+      return;
+    }
+    await this.secureStorageService.save(
+      `${options.userId}_${SecureStorageKeys.ldap}`,
```
MGibson1, line 56: What do these keys look like in the OSs secrets viewer (like mac's Keychain Access)?

Currently we see something like the attached image, where that item has a field, `account`, that holds identifying information.

Is that how this works?

image

MGibson1 (changes requested):
I don't see anything that would allow for multiple accounts, but I also don't see anything enforcing a single account.

Is multiple org sync from a single instance something we want to enable in the future?

What's the current experience if you set up your DC for an org, then log out and log into another org, set it up, then log out and log in to the first -- is your config maintained? Is that what we want?

differsthecat:
> I don't see anything that would allow for multiple accounts, but I also don't see anything enforcing a single account.
> 
> Is multiple org sync from a single instance something we want to enable in the future?
> 
> What's the current experience if you set up your DC for an org, then log out and log into another org, set it up, then log out and log in to the first -- is your config maintained? Is that what we want?

This isn't implementing multiple accounts in DC, it's just implementing the changes to state service.

MGibson1:
@differsthecat, we're still storing under some account-specific infrastructure. That's why I'm asking what would happen in that multi-user scenario I posed.

I'm wondering if we should wipe the state of all users if you log in as a new user

differsthecat:
@MGibson1 That's a good point. There could be some value in keeping the account settings when a different one is logged in, as the directory settings are then saved for the previous account. Only if that's a valid use case though.
Do you happen to know off hand if another account is logged in if the directory settings are persisted to the new account? If not, I agree and think we should just wipe the state if a new one is logged in.

src/services/state.service.ts
differsthecat: Yes, I just checked what it looks like for me and in the `account` field it has items such as `organization.{id}_ldapPassword` for the ldapPassword

src/services/state.service.ts
differsthecat: Agreed, this is something we probably want to do on the jslib stateService level as well. I'd like to get this PR in for now so that directory connector isn't blocked anymore but definitely an enhancement we can think about doing soon.

src/services/configuration.service.ts
```diff
@@ -1,64 +1,47 @@
+import { IConfiguration } from "src/models/IConfiguration";
```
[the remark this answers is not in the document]
differsthecat: Do you mind if we make this a tech debt item? I was just looking into doing this and we may have to do more than just deleting this service and dropping the directory items in there. I want to get this merged so people are unblocked and I'm afraid I might end up going down a rabbit hole doing this.

src/services/auth.service.ts
MGibson1: I don't believe we're doing anything that explicitly expects a guid vs a string in the future, but I believe making this consistent with other clients is a good idea.

Please make sure that the way this is implemented in Desktop is via a userId Guid (pretty sure it is) and change this to match.

src/services/state.service.ts
MGibson1: killer that's perfect

MGibson1:
Currently logout does not wipe settings, no. Logging in as another org also does not change directory settings.

We need to decide if we want to maintain that behavior or change it. I believe that right now we'd have unique Directory settings per organization login.

src/services/auth.service.ts
differsthecat: Just checked Desktop and it is using a guid value. I updated directory connector to match 👍

src/services/configuration.service.ts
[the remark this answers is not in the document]
differsthecat: This has been addressed with the latest commits!

src/services/stateMigration.service.ts
```diff
@@ -77,7 +77,7 @@ export class StateMigrationService extends BaseStateMigrationService {
   protected async migrateStateFrom1To2(useSecureStorageForSecrets: boolean = true): Promise<void> {
     await super.migrateStateFrom1To2();
     const state = await this.storageService.get<State<Account>>("state");
-    const userId = await this.storageService.get<string>(ClientKeys.clientId);
+    const userId = await this.storageService.get<string>("entityId");
```
MGibson1, line 80: Should we convert these strings to constants in either the `Keys` or `ClientKeys` hashes?

MGibson1 (changes requested):
One tiny little thing left. Sorry I missed this last time

src/services/stateMigration.service.ts
differsthecat: Yep, all of the keys in the migration service are now constants

MGibson1 (approved):

Issue #83, opened 2021-01-04
Azure Active Directory Sync Efficiency considerations


haneef95:
Hi,

I've been checking through the request logs for the Azure Active Directory from Bitwarden Directory Connector...

### Current settings
![image](https://user-images.githubusercontent.com/298208/103577871-7568d980-4ecd-11eb-803c-e2751ad4ce3f.png)
Sync Users with `includeGroup` filter
Sync Groups with no filters

### Efficiency measures
1. BWDC downloads all the users through `https://graph.microsoft.com/v1.0/users?$select=id,mail,userPrincipalName,displayName,accountEnabled`, however, it doesn't need a list of all the users, as it does not manifest in to the Bitwarden Organisation (due to the `includeGroup` User Filter)

2. Because of the includeGroup filter, BWDC keeps doing numerous `https://graph.microsoft.com/v1.0/users/{id}/checkMemberGroups` requests.

I guess it would be sufficient to just make use of either of the following for each 'includeGroup':
- `https://graph.microsoft.com/v1.0/groups/{groupId}/members?$select=id,mail,userPrincipalName,displayName,accountEnabled` - for direct members
- or `https://graph.microsoft.com/v1.0/groups/{groupId}/transitiveMembers?$select=id,mail,userPrincipalName,displayName,accountEnabled` - for nested members

3. BWDC downloads the members of every group, which does not manifest onto the Bitwarden Organisation, so, I believe that wouldn't be necessary either?

checkso:
This definitley needs to be addressed. 
In our tenant it takes ages to get all users.

haneef95:
I'm just thinking, it'll be a lot better if SCIM was implemented as part of the Bitwarden Public API. It has major benefits. AzureAD supports automatic provisioning through SCIM.

This has little votes, hopefully, it picks up: https://community.bitwarden.com/t/scim-as-an-alternative-to-directory-connector-for-user-management/28913

differsthecat:
Fixed 👍

haneef95:
Happy to see this in the release:

> Directory Connector - Azure AD Sync Performance: Performance for Directory Connector syncs against Azure Active Directory has been improved. Organizations syncing with Azure AD will not need to change their sync configuration.
https://bitwarden.com/help/releasenotes/#2022-02-08

PR #205, opened 2022-01-13
Only get users for includeGroup filter

differsthecat:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Azure AD gets all users for an entire directory regardless of the user filters provided, and then sends a separate api request to Azure per user to determine group membership to filter out unwanted users. This takes an unnecessarily long amount of time for clients that only want to sync a few users and have many many more users in their directory.

With this change, if the `includeGroup` filter is provided, only the users that belong to that group are synced over. This eliminates the need to check group membership on a per user basis.

The `excludeGroup` filter process has also been changed. We now get the users from those groups and use that list to determine whether to filter out the users instead of making a call to get a user's group membership.

I also left the process for getting deleted users the same because those functions rely on the delta token, which is not a supported by the Graph API for group members.

Even with those considerations, for an Azure AD with only 8 users, the sync time was reduced from an average of **2.3 seconds** down to **1 second**. This difference should be even more pronounced for larger instances that only sync over a select number of groups.

Ticket: https://app.asana.com/0/1201211497354246/1201456218559784
Issue: https://github.com/bitwarden/directory-connector/issues/83

## Code changes

- **src/services/azure-directory.service.ts:** 
    - Created a new method `getUsersByGroups`, that is invoked if the provided filter is `includeGroups`. This method calls the Microsoft Graph Api groups transitiveMembers route to get all members of a group, including nested members (members of groups that are in a group) Reference: https://docs.microsoft.com/en-us/graph/api/group-list-transitivemembers?view=graph-rest-1.0&tabs=http
## Testing requirements

Ensure that Azure AD syncs properly with all possible user filters. A lot of stuff changed here so this should be tested thoroughly. This could cause a large number of accounts to be unintentionally (de)provisioned which could cause some headaches.  It would be awesome to benchmark this against a larger Azure AD as well to see the full effects of this change.

## Before you submit

### DIFF
--- a/src/services/azure-directory.service.ts
+++ b/src/services/azure-directory.service.ts
@@ -86,31 +86,26 @@ export class AzureDirectoryService extends BaseDirectoryService implements IDire
   private async getCurrentUsers(): Promise<UserEntry[]> {
     const entryIds = new Set<string>();
     let entries: UserEntry[] = [];
+    let users: graphType.User[];
     const setFilter = this.createCustomUserSet(this.syncConfig.userFilter);
     const usersToExclude = new Set<string>();
 
     // Only get users for the groups provided in includeGroup filter
     if (setFilter != null && setFilter[0] === UserSetType.IncludeGroup) {
-      const users = await this.getUsersByGroups(setFilter);
-      if (users != null) {
-        entries = await this.buildUserEntries(users, usersToExclude, setFilter);
-      }
+      users = await this.getUsersByGroups(setFilter);
       // Get the users in the excludedGroups and filter them out from all users
     } else if (setFilter != null && setFilter[0] === UserSetType.ExcludeGroup) {
       (await this.getUsersByGroups(setFilter)).forEach((user: graphType.User) =>
         usersToExclude.add(user.id)
       );
       const userReq = this.client.api("/users" + UserSelectParams);
-      const users = await this.getUsersByResource(userReq);
-      if (users != null) {
-        entries = await this.buildUserEntries(users, usersToExclude, setFilter);
-      }
+      users = await this.getUsersByResource(userReq);
     } else {
       const userReq = this.client.api("/users" + UserSelectParams);
-      const users = await this.getUsersByResource(userReq);
-      if (users != null) {
-        entries = await this.buildUserEntries(users, usersToExclude, setFilter);
-      }
+      users = await this.getUsersByResource(userReq);
+    }
+    if (users != null) {
+      entries = await this.buildUserEntries(users, usersToExclude, setFilter);
     }
     return entries;
   }
@@ -136,8 +131,8 @@ export class AzureDirectoryService extends BaseDirectoryService implements IDire
     }
 
     const setFilter = this.createCustomUserSet(this.syncConfig.userFilter);
-    while (true) {
-      const users: graphType.User[] = res.value;
+    let users: graphType.User[] = res.value;
+    while (res[NextLink] != null) {
       if (users != null) {
         for (const user of users) {
           if (user.id == null || entryIds.has(user.id)) {
@@ -148,14 +143,13 @@ export class AzureDirectoryService extends BaseDirectoryService implements IDire
             continue;
           }
 
-          // Deleted users cannot be filtered by group in Azure AD so we only apply user filters if required
           if (
             setFilter != null &&
-            (setFilter[0] === UserSetType.IncludeUser || setFilter[0] === UserSetType.ExcludeUser)
+            (setFilter[0] === UserSetType.IncludeUser ||
+              setFilter[0] === UserSetType.ExcludeUser) &&
+            (await this.filterOutUserResult(setFilter, entry))
           ) {
-            if (await this.filterOutUserResult(setFilter, entry)) {
-              continue;
-            }
+            continue;
           }
 
           entries.push(entry);
@@ -171,6 +165,7 @@ export class AzureDirectoryService extends BaseDirectoryService implements IDire
       } else {
         const nextReq = this.client.api(res[NextLink]);
         res = await nextReq.get();
+        users = res.value;
       }
     }
 
@@ -385,11 +380,10 @@ export class AzureDirectoryService extends BaseDirectoryService implements IDire
 
       if (
         setFilter != null &&
-        (setFilter[0] === UserSetType.IncludeUser || setFilter[0] === UserSetType.ExcludeUser)
+        (setFilter[0] === UserSetType.IncludeUser || setFilter[0] === UserSetType.ExcludeUser) &&
+        (await this.filterOutUserResult(setFilter, entry))
       ) {
-        if (await this.filterOutUserResult(setFilter, entry)) {
-          continue;
-        }
+        continue;
       }
       if (!this.isInvalidUser(entry)) {
         entries.push(entry);
--- a/src/services/azure-directory.service.ts
+++ b/src/services/azure-directory.service.ts
@@ -148,7 +148,7 @@ export class AzureDirectoryService extends BaseDirectoryService implements IDire
             continue;
           }
 
-          // filterOutUserResult should only be called if the filter is includeUser or excludeUser
+          // Deleted users cannot be filtered by group in Azure AD so we only apply user filters if required
           if (
             setFilter != null &&
             (setFilter[0] === UserSetType.IncludeUser || setFilter[0] === UserSetType.ExcludeUser)
--- a/src/services/azure-directory.service.ts
+++ b/src/services/azure-directory.service.ts
@@ -86,21 +86,17 @@ export class AzureDirectoryService extends BaseDirectoryService implements IDire
   private async getCurrentUsers(): Promise<UserEntry[]> {
     const entryIds = new Set<string>();
     const entries: UserEntry[] = [];
-    const userReq = this.client.api("/users" + UserSelectParams);
-    let res = await userReq.get();
     const setFilter = this.createCustomUserSet(this.syncConfig.userFilter);
-    while (true) {
-      const users: graphType.User[] = res.value;
+
+    // Only get users for the groups provided in includeGroup filter
+    if (setFilter && setFilter[0] === UserSetType.IncludeGroup) {
+      const users = await this.getUsersByGroups(setFilter);
       if (users != null) {
         for (const user of users) {
           if (user.id == null || entryIds.has(user.id)) {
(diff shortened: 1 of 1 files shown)


src/services/azure-directory.service.ts
```diff
@@ -295,13 +317,10 @@ export class AzureDirectoryService extends BaseDirectoryService implements IDire
     const memberGroups = await this.client.api(`/users/${user.externalId}/checkMemberGroups`).post({
       groupIds: Array.from(setFilter[1]),
     });
-    if (memberGroups.value.length > 0 && setFilter[0] === UserSetType.IncludeGroup) {
-      return false;
```
eliykat, line 299: Maybe we should throw an error in this case, because this should never occur - and if it does, we probably want to abort syncing as this will invalidate our assumptions in this method.

src/services/azure-directory.service.ts
```diff
@@ -295,13 +317,10 @@ export class AzureDirectoryService extends BaseDirectoryService implements IDire
...
-    } else if (memberGroups.value.length > 0 && setFilter[0] === UserSetType.ExcludeGroup) {
+    if (memberGroups.value.length > 0 && setFilter[0] === UserSetType.ExcludeGroup) {
```
eliykat, line 320: You're checking the `UserSetType` in both `if` statements. Also, I don't think that we want to do this api call at all unless `setFilter[0] === UserSetType.ExcludeGroup`. I think you could move this check into the same `if` statement as `!checkGroupsFilter`, and just return early if it's not satisfied.

src/services/azure-directory.service.ts
```diff
@@ -367,6 +386,27 @@ export class AzureDirectoryService extends BaseDirectoryService implements IDire
...
 
+  private async getUsersByGroups(setFilter: [UserSetType, Set<string>]): Promise<graphType.User[]> {
+    const users: graphType.User[] = [];
+    for (const group of setFilter[1]) {
+      const groupUsersReq = this.client.api(
+        `/groups/${group}/transitiveMembers` + UserSelectParams
+      );
+      let res = await groupUsersReq.get();
+
+      while (true) {
```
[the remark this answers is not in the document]
eliykat: Same comment applies for when you move the other api call into its own method. `while(true)` loops seem bad.

eliykat (changes requested):
This is a good improvement, nice work @differsthecat!

Out of scope for this PR, but I wonder whether there is a better way of handling the `ExcludeGroup` case. e.g. get the transitive memberships of the excluded groups, get all users, then filter the users locally. The number of api calls would be roughly proportionate to the number of excluded groups rather than the number of users in the entire organization.

Please also flag this to QA especially, any accidental change in functionality could cause a large number of accounts to be unintentionally (de)provisioned which could cause some headaches.

src/services/azure-directory.service.ts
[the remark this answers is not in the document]
differsthecat: Agreed, I'll fix these!

src/services/azure-directory.service.ts
differsthecat: Yeah I removed these thinking they wouldn't be possible conditions anymore, but if I missed anything and it gets to here we should throw an error

src/services/azure-directory.service.ts
```diff
@@ -86,42 +86,64 @@ export class AzureDirectoryService extends BaseDirectoryService implements IDire
...
           ) {
             continue;
           }
-
           entries.push(entry);
           entryIds.add(user.id);
         }
       }
+    } else {
+      const userReq = this.client.api("/users" + UserSelectParams);
```
differsthecat, line 113: Updated! There's a slight bit of duplication of some minor things but I feel like it is much more readable now

src/services/azure-directory.service.ts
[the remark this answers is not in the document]
differsthecat: Refactored this code into the `getUsersByResource()` method so there are no more `while(true)`'s.
Also I am open to a better name for that method, that was the best I could come up with 😅

src/services/azure-directory.service.ts
```diff
@@ -86,42 +86,64 @@ export class AzureDirectoryService extends BaseDirectoryService implements IDire
...
+            if (await this.filterOutUserResult(setFilter, entry, true)) {
+              continue;
+            }
 
-      if (res[NextLink] == null) {
-        break;
-      } else {
-        const nextReq = this.client.api(res[NextLink]);
-        res = await nextReq.get();
+            if (
```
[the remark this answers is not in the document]
differsthecat: Extracted into `isInvalidUser()`

differsthecat:
@eliykat Thank you for the thorough review! I accidentally deleted one of your comments when I went to delete one of mine :face-palm: but I was able to implement a lot of fixes here! Also awesome idea about how to handle `excludeGroups`, I went ahead and implemented it here! With that implemented there wasn't even a need to keep the code that sent requests via `checkMemberGroups`

src/services/azure-directory.service.ts
eliykat: Reminder to do this if it's called with a group filter. Will be a good safeguard in case it's called from elsewhere in the code, or if someone else misses the refactor.

eliykat (changes requested):
Looks great! Just one thing that I think you missed.

src/services/azure-directory.service.ts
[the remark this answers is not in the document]
eliykat: I think the name is good! I like that you've abstracted the api call to that one method.

src/services/azure-directory.service.ts
eliykat: I think you may be able to reduce this duplication further. Each `for (const user of users)` loop basically does the same thing, but with a different filter:
* IncludeGroup - no filter (filtering is done in the api call)
* ExcludeGroup - not in `usersToExclude`
* users filters - `filterOutUserResult` call

So maybe you can move this loop to a method that takes a filter callback or something similar.

eliykat (commented):
Had a bit more of a think about this as well.

differsthecat:
> Had a bit more of a think about this as well.

Implemented this!

src/services/azure-directory.service.ts
```diff
@@ -157,8 +147,15 @@ export class AzureDirectoryService extends BaseDirectoryService implements IDire
           if (!entry.deleted) {
             continue;
           }
-          if (await this.filterOutUserResult(setFilter, entry, false)) {
-            continue;
+
+          // filterOutUserResult should only be called if the filter is includeUser or excludeUser
```
eliykat, line 151: This just restates the `if` condition, I was thinking of something more like

```suggestion
          // Deleted users cannot be filtered by group in Azure AD so we only apply user filters if required
```

...assuming that is correct, that's my understanding of the rationale. Basically replacing the old comment on line 291 that we're now deleting.

src/services/azure-directory.service.ts
```diff
@@ -157,8 +147,15 @@ export class AzureDirectoryService extends BaseDirectoryService implements IDire
...
+          if (
+            setFilter != null &&
+            (setFilter[0] === UserSetType.IncludeUser || setFilter[0] === UserSetType.ExcludeUser)
+          ) {
+            if (await this.filterOutUserResult(setFilter, entry)) {
```
eliykat, line 156: This can be combined into the previous if condition

src/services/azure-directory.service.ts
```diff
@@ -85,43 +85,33 @@ export class AzureDirectoryService extends BaseDirectoryService implements IDire
...
-          }
+    const usersToExclude = new Set<string>();
 
-          entries.push(entry);
-          entryIds.add(user.id);
-        }
+    // Only get users for the groups provided in includeGroup filter
+    if (setFilter != null && setFilter[0] === UserSetType.IncludeGroup) {
+      const users = await this.getUsersByGroups(setFilter);
+      if (users != null) {
```
eliykat, line 95: This `if` block is identical in each branch. You can extract it to the bottom of the `if` condition.

src/services/azure-directory.service.ts
```diff
@@ -367,6 +346,63 @@ export class AzureDirectoryService extends BaseDirectoryService implements IDire
...
+        `/groups/${group}/transitiveMembers` + UserSelectParams
+      );
+      users.push(...(await this.getUsersByResource(groupUsersReq)));
+    }
+    return users;
+  }
+
+  private async buildUserEntries(
+    users: graphType.User[],
+    usersToExclude: Set<string>,
```
eliykat, line 374: Suggest renaming this to `userIdsToExclude` just so it's clear what the `string` is.

src/services/azure-directory.service.ts
eliykat: We decided not to do this so all good

eliykat (changes requested):
A few minor comments then good to go!

src/services/azure-directory.service.ts
differsthecat: Done

src/services/azure-directory.service.ts
differsthecat: Fixed; this is the same condition we did for the deleted users, so I went ahead and removed the comment as it seems a bit unnecessary since it's the same here too.

src/services/azure-directory.service.ts
differsthecat: Fixed; this is the same condition we did for the deleted users, so I went ahead and removed the comment as it seems a bit unnecessary since it's the same here too.

src/services/azure-directory.service.ts
```diff
@@ -136,8 +131,8 @@ export class AzureDirectoryService extends BaseDirectoryService implements IDire
     }
 
     const setFilter = this.createCustomUserSet(this.syncConfig.userFilter);
-    while (true) {
```
eliykat, line 149: I like trying to replace the `while (true)` loop, but there's not much point unless you can also refactor the break loop logic on line 160. Currently you're still breaking out of the loop with the `break` statement rather than the `while` condition.

eliykat (changes requested):

src/services/azure-directory.service.ts
differsthecat: Gotcha, I was thinking this way it was still easier to understand. I put it back the way it was.

eliykat (approved):

PR #214, opened 2022-01-26
[bug] Remove redundant state clean call

addisonbeck:
https://app.asana.com/0/1201648796371593/1201694099230961

## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Bug
When logging out the process succeeds, but an error is logged to the console about not being able to clear tokens. 

## Code changes
* We are redundantly cleaning state twice right now, once in the [main logout function](https://github.com/bitwarden/directory-connector/blob/9e3d1caee4c8f986fb6acff0cff7fc2375c53491/src/bwdc.ts#L262-L265), and once in the DC extended auth service. This is resulting in the method being called twice for logouts, and the second time throws an error because it expects to have something to clean up. I removed the auth service call, though I think removing either would have the same effect.

## Before you submit

### DIFF
--- a/src/services/auth.service.ts
+++ b/src/services/auth.service.ts
@@ -66,10 +66,6 @@ export class AuthService extends AuthServiceBase {
     return await super.logInApiKey(clientId, clientSecret);
   }
 
-  async logOut(callback: Function) {
-    super.logOut(callback);
-  }
-
   private async organizationLogInHelper(clientId: string, clientSecret: string) {
     const appId = await this.appIdService.getAppId();
     const entityId = clientId.split("organization.")[1];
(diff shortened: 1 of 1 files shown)


src/services/auth.service.ts
```diff
@@ -67,7 +67,6 @@ export class AuthService extends AuthServiceBase {
   }
 
   async logOut(callback: Function) {
-    this.stateService.clean();
```
vincentsalucci, line 70: This `logOut` superclass'd method is only called from one entry point in `dc`. Removing it here seems fine. Would that also mean you can remove the entire method as well?

addisonbeck: Good point, Vince, it does seem safe to just remove the whole logOut override. In the code for last release it just called a method to clear apiKeys, which is happening in stateService.clean, and so is happening in the bwdc.logout() method. 

I update this to remove this method.

vincentsalucci (approved):

PR #245, opened 2022-03-04
Use saveAccount to scaffold new account

eliykat:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix the following defect: when upgrading from the current release of d-c, logging in fails silently.

Asana: https://app.asana.com/0/1169444489336079/1201909137219813/f

## Code changes

 

d-c has its own account scaffolding method in `scaffoldNewAccountDiskStorage`, but it was calling `storageService.save` directly instead of using `saveAccount` in the base class. You must use `saveAccount`, because that deals with any stale disk cache (see https://github.com/bitwarden/jslib/pull/668). Otherwise, it writes the new value to disk but does not clear the old disk cache, and `stateService` methods continue reading the old cached values.

## Testing requirements

Regression testing on logging in:
* after upgrading
* on a fresh install (or after clearing data.json)
* in the gui and cli

## Before you submit

### DIFF
--- a/src/services/state.service.ts
+++ b/src/services/state.service.ts
@@ -529,12 +529,12 @@ export class StateService
   }
 
   protected async scaffoldNewAccountDiskStorage(account: Account): Promise<void> {
-    const storedAccount = await this.getAccount(
-      this.reconcileOptions(
-        { userId: account.profile.userId },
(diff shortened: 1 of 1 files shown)


differsthecat (approved):
Looks good!

PR #266, opened 2022-04-29
[EC-176] Fix CLI errors caused by server URLs

differsthecat:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

[Context](https://bitwarden.atlassian.net/browse/EC-176)
Fixes syncing from the CLI would throw the error `Cannot read properties of null (reading 'directorySettings')` for some users. Determined that this was happening for people with server url's configured and that the CLI was getting the base url on the account level. We are never going to utilize separate accounts in the Directory Connector, override stateService.getEnvironmentUrls to only return the global ones. 

Also a refactor to `stateService` for how we handle saving the account, given that d-c is a bit different to the other clients.

## Testing requirements
See linked ticket
## Before you submit

### DIFF
--- a/src/services/state.service.ts
+++ b/src/services/state.service.ts
@@ -53,6 +53,34 @@ export class StateService
     super(storageService, secureStorageService, logService, stateMigrationService, stateFactory);
   }
 
+  async saveAccountToDisk(account: Account, options: StorageOptions): Promise<void> {
+    const accountNoSecrets = this.removeSecrets(account);
+    return super.saveAccountToDisk(accountNoSecrets, options);
+  }
+
+  private removeSecrets(account: Account): Account {
+    if (this.useSecureStorageForSecrets) {
+      const configs = account.directoryConfigurations;
+      if (configs.azure != null) {
+        configs.azure.key = StoredSecurely;
+      }
+      if (configs.gsuite != null) {
+        configs.gsuite.privateKey = StoredSecurely;
+      }
+      if (configs.ldap != null) {
+        configs.ldap.password = StoredSecurely;
+      }
+      if (configs.okta != null) {
+        configs.okta.token = StoredSecurely;
+      }
+      if (configs.oneLogin != null) {
+        configs.oneLogin.clientSecret = StoredSecurely;
+      }
+    }
+
+    return account;
+  }
+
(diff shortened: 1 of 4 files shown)


src/services/state.service.ts
```diff
@@ -53,6 +54,61 @@ export class StateService
...
   }
 
+  async saveAccountToDisk(account: Account, options: StorageOptions): Promise<void> {
+    await this.setSecureSecrets(account.directoryConfigurations);
+    const accountNoSecrets = this.removeSecrets(account);
+    return super.saveAccountToDisk(accountNoSecrets, options);
+  }
+
+  private removeSecrets(account: Account): Account {
+    if (this.useSecureStorageForSecrets) {
```
eliykat, line 64: The problem I was having last week is that this will also remove secrets from the in-memory account object.

I can think of 2 ways around this:
1. Create a copy of the account object using `Object.assign` (call it `savedAccount`), then create a copy of the directoryConfig object, add it to the `savedAccount` object, and remove secrets from that. Seems messy but would leave the in memory account object (and its directoryConfig) intact.
2. Remove directoryConfig from the account object. Although then we hit migration problems, may not be worth it.

src/services/state.service.ts
```diff
@@ -53,6 +54,61 @@ export class StateService
...
+      }
+      if (configs.oneLogin != null) {
+        configs.oneLogin.clientSecret = StoredSecurely;
+      }
+    }
+
+    return account;
+  }
+
+  private async setSecureSecrets(config: DirectoryConfigurations) {
```
eliykat, line 86: This seems like it replicates `setDirectory`, do we need it as a separate method?

src/services/state.service.ts
```diff
@@ -568,4 +608,8 @@ export class StateService
     };
     return Object.assign(this.createAccount(), persistentAccountInformation);
   }
+
+  async getEnvironmentUrls(options?: StorageOptions): Promise<EnvironmentUrls> {
```
eliykat, line 612: Can't believe the fix was this simple 😂  good catch

eliykat (changes requested):

PR #269, opened 2022-05-03
[EC-176] Fix CLI errors caused by server URLs

differsthecat:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

[Context](https://bitwarden.atlassian.net/browse/EC-176)
Fixes syncing from the CLI would throw the error Cannot read properties of null (reading 'directorySettings') for some users. Determined that this was happening for people with server url's configured and that the CLI was getting the base url on the account level. We are never going to utilize separate accounts in the Directory Connector, override stateService.getEnvironmentUrls to only return the global ones.

## Code changes

- **src/services/state.service.ts:** Override state service to only use global environment urls

## Before you submit

### DIFF
--- a/src/services/state.service.ts
+++ b/src/services/state.service.ts
@@ -2,6 +2,7 @@ import { LogService } from "jslib-common/abstractions/log.service";
 import { StateMigrationService } from "jslib-common/abstractions/stateMigration.service";
 import { StorageService } from "jslib-common/abstractions/storage.service";
 import { StateFactory } from "jslib-common/factories/stateFactory";
+import { EnvironmentUrls } from "jslib-common/models/domain/environmentUrls";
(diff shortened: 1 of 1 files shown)


src/services/state.service.ts
```diff
@@ -568,4 +569,8 @@ export class StateService
     };
     return Object.assign(this.createAccount(), persistentAccountInformation);
   }
+
+  async getEnvironmentUrls(options?: StorageOptions): Promise<EnvironmentUrls> {
+    return await this.getGlobalEnvironmentUrls(options);
```
eliykat, line 574: `await` is unnecessary here, `getEnvironmentUrls` will already be `await`ed by whoever calls it

eliykat (changes requested):

eliykat (approved):

PR #286, opened 2022-10-21
[EC-488] Handle falsy values in keytarSecureStorageService

eliykat:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix the following bug:

1. log in to bwdc CLI
2. run `bwdc config directory 0`
3. observe error: `Password is required.`

This error is being thrown by keytar - "Password" is referring to the secret being saved in the keychain (not the user's password/api key).

Keytar throws if you try to save a falsy value. The config command basically loads and then re-saves all config settings, so any secrets that aren't explicitly set will be saved to the keychain as null values, triggering this error.

This was introduced with account switching. Before that, `setDirectory` checked for `null` and would call `secureStorageService.remove` instead. That check was removed.

## Code changes

- **src/services/keytarSecureStorage.service.ts** - check for null and call `remove` instead. By implementing it here we avoid this problem regardless of the caller, and this method can never be validly called with a null value anyway.

## Before you submit

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -101,6 +101,7 @@ jobs:
 
       - name: Version Test
         run: |
+          sudo apt-get update
           sudo apt install libsecret-1-0 dbus-x11 gnome-keyring
           eval $(dbus-launch --sh-syntax)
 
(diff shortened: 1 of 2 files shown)


r-tome (dismissed):

eliykat:
I didn't sign this commit the first time, so I ran `git reset master`, re-committed the same changes, and force pushed. No actual change to the diff.

eliykat:
The build pipeline is failing because a remote file is missing. I've messaged devops about it.

r-tome (approved):

PR #333, opened 2023-06-07
JS library Angular and Jest updates

withinfocus:
## Type of change

- [ ] Bug fix
- [ ] New feature development
- [X] Tech debt (refactoring, code cleanup, dependency upgrades, etc)
- [ ] Build/deploy pipeline (DevOps)
- [ ] Other

## Objective

Few manual upgrades given that Renovate is getting annoyed with artifact updates.

## Code changes

- **jslib/package.json:** Angular and Jest.
- **jslib/package-lock.json:** Lockfile (forced update).

## Before you submit

- [ ] I have checked for **linting** errors (`npm run lint`) (required)
- [ ] I have added **unit tests** where it makes sense to do so (encouraged but not required)
- [ ] This change requires a **documentation update** (notify the documentation team)
- [ ] This change has particular **deployment requirements** (notify the DevOps team)

PR #335, opened 2023-06-10
Move JS library dependencies up to root

withinfocus:
## Type of change

- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Relates to #333. Moves dependencies and commands up to the root for easier maintenance.

## Before you submit


Hinton (approved):
LGTM. There are some submodule related scripts in the package.json that we can remove at a later stage. We should probably also look into adding a jest config to the root so we can run tests in the clients.

PR #352, opened 2023-08-31
[AC-1614] Updated webpack config

r-tome:
## Type of change

- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Directory-connector was upgraded to Angular v15 and the electron build has stopped working since.
This PR fixes that by adding a missing webpack config.

## Code changes

- **webpack.renderer.js:** Added required webpack configuration
- **jslib/angular/src/components/callout.component.html:** Ran prettier
- **jslib/common/src/enums/importOptions.ts:** Ran prettier
- **src/app/tabs/dashboard.component.html:** Ran prettier
- **src/app/tabs/more.component.html:** Ran prettier

## Testing requirements

Test that both the CLI and the App work as expected.

## Before you submit

- [X] I have checked for **linting** errors (`npm run lint`) (required)

### DIFF
--- a/jslib/angular/src/components/callout.component.html
+++ b/jslib/angular/src/components/callout.component.html
@@ -12,10 +12,10 @@
     {{ enforcedPolicyMessage }}
     <ul>
       <li *ngIf="enforcedPolicyOptions?.minComplexity > 0">
-        {{ "policyInEffectMinComplexity" | i18n: getPasswordScoreAlertDisplay() }}
+        {{ "policyInEffectMinComplexity" | i18n : getPasswordScoreAlertDisplay() }}
       </li>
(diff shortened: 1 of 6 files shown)


eliykat (approved):
Looks good to me, thanks for investigating this!

My suggestion is that this is merged without QA, given that `master` is totally broken without it, then QA can regression test everything after you've done the dependency upgrade ticket. Save them testing it twice. Please double check with them first though.

r-tome:
@eliykat thanks for reviewing. The dependency upgrades have already been done so by testing this branch we'll be testing those changes also.

vincentsalucci (approved):

PR #374, opened 2023-12-07
[AC-1743] pt. 1: Unpackage-ify jslib

addisonbeck:
# [\[AC-1743\]](https://bitwarden.atlassian.net/browse/AC-1743) pt. 1 ⮕ Unpackage-ify jslib

## Type of change

- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

As part of an initiative to upgrade dependencies across this project it has become necessary to take a few steps toward better integrating the old jslib code from pre-monorepo days.

### 🚩 The Problem

Phase one of this work has been to identify what packages in `package.json` are unused by the project. To help with this a tool called [depcheck](https://github.com/depcheck/depcheck) can be used that will scan source code and provide a list of unused dependencies. This tool does not work on Directory Connector as it stands. Both Directory Connector and depcheck consider each of the `jslib` projects to be their own separate projects external to Directory Connector. When depcheck is run it reports *all* dependencies as unused. This happens because dependencies are managed in the `package.json` in the project root, but there are many other empty `package.json` files scattered throughout the source code to provide arbitrary separation of these "packages". This configuration is outdated, as jslib is no longer an external dependency and instead is fully managed as part of the Directory Connector project.

### 🛠️ The Solution

Jslib needs to be unpackage-ified and better merged into the Directory Connector package configuration. This PR takes some steps to get there, but stops once the needs for successfully running depcheck against the project are met. There is more work to be done in this area at a later time, and some proposed steps to be taken are listed below. For now the following updates have been made:

1. On [`f3c4157` Unpackage-ify jslib](https://github.com/bitwarden/directory-connector/commit/f3c4157ac471e80dbc7af028b2f2667741d3e9d9) the project was restructured to only have one `package.json` file that lives in root. This allows depcheck to consider all the source code to be one project and analyze its dependency usage. This involved:
    1. All package.json files in the jslib subproject directories have been deleted. These files were mostly empty and carried no consequential configuration.
    1. The package.json under `/src/` has been deleted. This file contained a version number, a product name, and a reference to `main.js`. These configuration items have been moved up to the root `package.json`, and CI workflows have been updated to reference it instead.
    1. `jslib` package references were removed from `package.json`.
1. On [`42dbf0d` Adjust .tsconfig path for root and apply to jslib](https://github.com/bitwarden/directory-connector/commit/42dbf0d884be9bc4e71216de14471e4f7e9a0a58) The `.tsconfig` paths for jslib were removed in favor of one path that points to project root. This does a few things:
    1. Stops depcheck from considering jslib imports to be "missing" dependencies. They don't support .tsconfig paths [as a rule](https://github.com/depcheck/depcheck/issues/564#issuecomment-778634977).
    1. Moves away from using paths to manage monorepo-like imports. Directory connector being a standalone app doesn't align with the current project layout of having jslib being externally referenced.
    1. Pleases me on a personal level. I'm not a huge fan of tsconfig paths myself, though obviously this is an opinion and open to discussion.
1. [`35ae96a` Rebuild package-lock.json](https://github.com/bitwarden/directory-connector/commit/35ae96a0602f371e2ac39de16edcc087cb242e30) Rebuilds `package-lock.json`
1. [`4f2c986` Disable husky in CI](https://github.com/bitwarden/directory-connector/commit/4f2c9860c15e6c24dacde1ebfed3d8b3c00151cf) Stops husky from running in CI. Admittedly, I'm not sure what happened here - but I started getting build failure errors in CI because husky couldn't find the `.git` folder after these changes. We don't need husky to run in CI anyway, so it seemed safe to just turn this off.

## ⏭️ What's next?

* I'd like to create the following tech debt items to continue reworking jslib into the project. Please provide feedback on these if you have any, and I'll be creating them after the PR is approved.
    - [ ] Move jslib into the `src` folder
    - [ ] Identify and remove unused jslib code from the project
    - [ ] Merge the remaining jslib code into the natural directory structure of Directory Connector
    - [ ] Compare `directory-connector` to `bitwarden/template` for any useful tools to add to the project
* Next up in this dependency upgrade effort I'll be:
    1. Removing a bunch of unused dependencies
    1. Taking another look at the value of the test setup in the project and seeing what can be saved
    1. Fixing eslint in the project and accounting for its plugins
    1. Performing any patch upgrades for outdated packages
    1. Performing any minor upgrades for outdated packages
    1. Performing any "safe" major upgrades for outdated packages
    1. Upgrading Angular, Webpack, Node, and npm to their latest LTS versions

## Testing requirements

This is really just a code & config restructure and should not have any impact on functionality.

## Before you submit

[AC-1743]: https://bitwarden.atlassian.net/browse/AC-1743?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package-lock.json
eliykat, line 1: It looks like all our dependencies in `package.json` are already pinned, so regenerating this should be safe.

eliykat (commented):
Thanks for tackling this @addisonbeck! 

This looks good to me, however @Hinton would you mind looking over the approach here as you have experience structuring & packaging electron apps. No need to review the code, just any general feedback you might have.

## Local testing

I tried to fire it up locally just to verify myself, but I got this error:

Screenshot 2023-12-11 at 2 00 53 pm

I'm sure I've seen it before, but not in a long time. Any suggestions?

The build artifacts seem to work OK though.

## Feedback

> The package.json under /src/ has been deleted. 

I thought this was required for packaging reasons, but apparently [electron-builder](https://www.electron.build/tutorials/two-package-structure.html) no longer requires it. So this seems fine, just wanted to note it.

> order of operations

It looks like you're going to cull dependencies first, then remove dead code later in a separate follow-up ticket. Can depcheck detect when a dependency is only used by dead code? If yes and you remove those dependencies, won't that cause a bunch of errors in the dead-but-not-yet-removed code?

I'm OK with whatever works - just asking the question.

> Removing tests from the project. This is rather sad, but they don't work and I have had a hard time getting them working again. Most of the tests aren't relevant to Directory Connector anyway, and we should start from scratch there. I'll be creating a tech debt item to address this on a later PR.

Any tests for dead code should be removed along with the dead code, but if we're relying on the code then we should keep the tests we have, right? Even if it's not d-c specific, we can't assume that it's tested in `clients` anymore as the two repos have diverged. What issues are you encountering with getting the tests to run?

> Moves away from using paths to manage monorepo-like imports. Directory connector being a standalone app doesn't align with the current project layout of having jslib being externally referenced.

I like tsconfig paths in `clients` more than you do, but I agree that d-c doesn't need them. For an app this size, we should just have the files we need in the `src` directory (and there shouldn't be that much in there once dead code is removed).

> other tasks

As a follow-up task (after dependency upgrades), you could compare this repo to https://github.com/bitwarden/template/ which contains template tooling, config, etc, to see if there's anything worth bringing over. Similarly, compare to `clients` as our major TS repo to consider poaching eslint/prettier configs etc. Ideally we should have the same configs to enforce the same conventions (where applicable).

Hinton:
I haven't reviewed the PR just addressing the general approach per @eliykat request.

The repository contains two projects, the CLI and the DC desktop app. As such I would still consider this a monorepository. Wile we technically don't need the libs `package.json` files due to the way we build our application, tests and possibly other things rely on it to my knowledge and for that reason I would be a bit hesitant in removing them.

I suspect you could get rid of almost all code changes by keeping the existing `tsconfig` aliases which would decrease the changelog in this PR, but changing to `@` should also work.

I've never used the single package.json config for electron-builder, and due to the repository not being a standard electron app I'm unsure how it would interact with native dependencies like `keytar`. Historically we've declared native dependencies in the `src/package.json`.

Deleting all tests seems a bit odd, I would be hesitant to remove used tests since there is a high probability we won't bring them back.

addisonbeck:
## [re: removing tests]

I agree that removing tests is unpleasant. I'm not a jest expert, but I was unable to get any of them to run correctly even on `main`. This issue, the knowledge that many of the tests are unrelated to Directory Connector, and the outdated-ness of Directory Connector's test configuration compared to `clients` all contributed to the idea to remove the tests. I'll spend some time seeing what percentage of the existing tests are relevant to DC itself (instead of other clients from when this was jslib) and come back with more information. I updated the task list in the PR description to match this.

## [re: comments from @eliykat]

> I tried to fire it up locally just to verify myself, but I got this error:

I have not seen this, but I'll see if I can recreate it. In the meantime: do you have any lingering `node_modules` folders about? If you've got one under `/src/` or didn't  rebuild them in root that may be related.

>  Is depcheck smart enough to mark deps only used by dead code as unused?

No. Depcheck checks for imports or import-like statements, but doesn't go much deeper than that to my knowledge. We would want to check for unused dependencies again after any dead code is removed. There are still quite a few unused dependencies as it stands. A quick scan of my notes tells me that I removed 11 packages in my dev branch, though this could actually be a bit lower or higher. I think removing dead code from jslib is a separate body of work that would just include taking another look to see if we unearthed anymore unused dependencies. 

> other tasks

Okay, I updated the tech debt task list in this PR description to include taking a look at `bitwarden/template`. I'm already porting eslint and prettier configs over from `clients` in the next phase or two of this work, so that will already be taken care of.

## [re: comments from @Hinton]

> As such I would still consider this a monorepository. 

I think this is tricky. DC and the DC CLI certainly are two independently runnable programs, but they don't exist separately in the project structure by any means as it stands right now.  [`program.ts`](https://github.com/bitwarden/directory-connector/blob/master/src/program.ts) is even in the Angular apps `/src/` folder. I'm worried we will end up dramatically increasing the scope of this work if we try and consider the two as separate at this stage.

>  tests and possibly other things rely on [the `jslib/package.json` files] to my knowledge and for that reason I would be a bit hesitant in removing them.

This did not appear to be the case to me. The `package.json` files in `jslib` didn't have any meaning configuration in them - and the tests for Directory Connector don't work on `master`.

>  I suspect you could get rid of almost all code changes by keeping the existing tsconfig aliases

You're correct. The path changes just keep us from having to ignore jslib when running the dependency check tool.

> I've never used the single package.json config for electron-builder, and due to the repository not being a standard electron app I'm unsure how it would interact with native dependencies like keytar

I was not aware of this quirk with electron, and that makes the way this was set up previously make more sense. It seems based on @eliykat's comment above that this is no longer an issue, though.  I've also had no issue interacting with keytar during development. Authentication passwords for syncing directories are saved there and are coming back fine. I feel okay just making a note to have QA check for issues in this area.

jslib/common/src/misc/sequentialize.ts
```diff
@@ -6,7 +6,7 @@
  *
  * Results are not cached, once the promise has returned, the next call will result in a fresh call
  *
- * Read more at https://github.com/bitwarden/jslib/pull/7
+ * Read more at https://github.com/bitwarden/@/jslib/pull/7
```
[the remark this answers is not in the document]
addisonbeck: Proof accepted. Nice catch! Fixed on https://github.com/bitwarden/directory-connector/pull/374/commits/64e4082099917f40f4309551e495b794d71809dc.

.eslintrc.json
```diff
@@ -4,7 +4,6 @@
     "browser": true,
     "node": true
   },
-  "extends": ["./jslib/shared/eslintrc.json"],
```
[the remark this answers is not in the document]
addisonbeck: Good call. Addressed on https://github.com/bitwarden/directory-connector/pull/374/commits/c526cb275b7a6da3ee70951231ee822ff492d0b9.

package.json
```diff
@@ -188,6 +186,6 @@
   },
   "lint-staged": {
     "./!(jslib)**": "prettier --ignore-unknown --write",
-    "*.ts": "eslint --fix"
+    "./**!(.spec).ts": "eslint --fix"
```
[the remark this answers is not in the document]
addisonbeck: I think I made this change when I was panicking trying to figure out why lint checks were blowing up like crazy, and never put it back after I realized Directory Connector just has a lot of lint errors. This change has been reverted on https://github.com/bitwarden/directory-connector/pull/374/commits/3037c29e2e3f68bd9f3d1f2c69525ceddd32084c.

Hinton:
I've resolved the jest test suite not working in https://github.com/bitwarden/directory-connector/pull/379. Turns out it got left behind when Angular was upgraded.

addisonbeck:
Nice! #379 is approved. That clears that up. Thanks @Hinton!

joseph-flinn:
Heyo! I heard that builds are hit or miss. There are some edge cases in the triggers in `directory-connector` current approach to CI triggers. Setting the `on` triggers in `.github/workflows/build.yml` to the following will fix it:

```yaml
---
name: Build

on:
  pull_request: {}
  workflow_dispatch: {}
```

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@bitwarden/directory-connector",
   "productName": "Bitwarden Directory Connector",
   "description": "Sync your user directory to your Bitwarden organization.",
-  "version": "2023.10.0",
+  "version": "2023.12.0",
```
eliykat, line 5: There was a mistaken version bump on `main` which was just reverted. We should still be at 2023.10.0.

package-lock.json
eliykat, line 1: There are actually some minor version bumps in this file, any idea why that is? Maybe dependencies-of-dependencies?

Nothing jumps out at me as a problem, just mentioning for visibility.

eliykat (changes requested):
Just a few minor things and then we should be good to go! Feel free to queue up that next PR as well.

package.json
addisonbeck: 👍 https://github.com/bitwarden/directory-connector/pull/374/commits/a97c6f61cc735fe9a1ebcfd98f1184df83857cf5

eliykat (approved):

eliykat:
Good work! You may just need to chase up a review from @bitwarden/dept-devops for the workflow changes.

vgrassia (approved):

PR #393, opened 2023-12-19
[AC-1743] pt. 2: Update eslintrc and fix any errors

addisonbeck:
# [\[AC-1743\]](https://bitwarden.atlassian.net/browse/AC-1743) pt. 2 ⮕ Update eslintrc and fix any errors

## Type of change

- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Previous parts in this series

* https://github.com/bitwarden/directory-connector/pull/374

## Objective

As part of an initiative to audit dependencies across this project it has become necessary to take a look at how eslint is configured in the project compared to the standard set by the `clients` monorepo.

### 🚩 The Problem

The `eslint` config in `clients` and what exists in Directory Connector have gotten a bit out of sync. This makes it difficult to tell what plugins in use now we want to upgrade, or if there are any we can discard.

### 🛠️ The Solution

Lets do a refresh of Directory Connector's `eslintrc` when compared to the `clients` monorepo so we have a clean slate for auditing any plugins in the next phase of this work. The commits in this PR do the following:

1. Lift the `eslintrc` from `clients` and slim it down to meet the needs relevant to Directory Connector. 
    * This involved adding a few new plugins 
1. Fix any errors or warnings that came up during the process
1. Running `prettier`

## ⏭️ What's next?

Next up in this dependency upgrade effort I'll be:

1. Removing a bunch of unused dependencies
1. Performing any patch upgrades for outdated packages
1. Performing any minor upgrades for outdated packages
1. Performing any "safe" major upgrades for outdated packages
1. Upgrading Angular, Webpack, Node, and npm to their latest LTS versions

## Screenshots

https://github.com/bitwarden/directory-connector/assets/15897251/5cf9f7d9-8712-4338-935a-ba5d438d9756

## Testing requirements

This is really just a code & config restructure and should not have any impact on functionality.

## Before you submit


jslib/common/src/abstractions/state.service.ts
eliykat, line 1: This is a meaningful change, `BehaviorSubject`s and `Observable`s are different things.

Assuming it's an eslint error, it's better to disable it for the offending line using the comment syntax: https://eslint.org/docs/latest/use/configure/rules#using-configuration-comments-1

I say this because we don't actually want to make semantic changes at this point, we're just trying to get eslint up-to-date and passing. It's not worth triggering QA and potentially causing regressions for.

The minor syntax changes are fine (e.g. the location of curly braces in import statements), it's mostly the rxjs stuff I'd rather leave alone.

eliykat (changes requested):
I've compared eslint config to `clients`, it all looks good and sensible, just 1 request over some of the eslint fixes.

Also, I didn't realise we don't have linting running in CI 😬

jslib/common/src/abstractions/state.service.ts
addisonbeck: Ahhh, I don't know. I'd like to ask you to reconsider. Here are some excuses as to why I'd say this is safe to do:

1. The underlying `BehaviorSubject` is not going anywhere. It's still the backbone of how `StateService` works, and is still what is referenced privately in the `StateService`, which is where most of the magic happens. The direct `accounts` value was only referenced directly from outside of `StateService` in two places: `EnvironmentService` [here](https://github.com/bitwarden/directory-connector/pull/393/files#diff-6a56b3fe6e7835ff35cfba470ea568970acf7a8330c48b63e446ff1007c8b4fdR24-R32) and `VaultTimeoutService` [here](https://github.com/bitwarden/directory-connector/pull/393/files#diff-d991c28464a81aa68fadf4c6fd2d9b01b451d3b33aa6699893c4c764c163a88fR72-R73). From what I can tell Directory Connector doesn't use `VaultTimeoutService` at all other than bringing it along for the ride from `JslibServicesModule`. There isn't any activity tracking or handling of a lock action at all. 
2. The way these two outside references to the `accounts` `BehaviorSubject` had to be changed to work with `Observables` instead was minimal. The `VaultTimeout` example was calling `getValue()` and only changed to a `firstValueFrom`. Very simple, and again, this method doesn't do anything in DC to my knowledge other than maybe send a message that is never handled by the client. The `EnvironmentService` example is a bit more complicated in the change, but it still just ends in a simple `subscribe`. This is the exact same way this code works in `clients` as of [right now](https://github.com/bitwarden/clients/blob/062a3ce2d2a33d156bde009cae8778db29060ba2/libs/common/src/platform/services/environment.service.ts#L53-L64)
3. Account Switching does not exist as a concept in Directory Connector, meaning both of these changes are only triggered on initial login. With that in mind there isn't much to test here other than that setting environment urls still works, which I would assume will be covered in the regression run that we are triggering at the end of this work (though I will make an explicit note about this in the notes for QA). 

Did I write all of this just to not have to disable eslint on a few lines of code that likely won't even make the cut when we do a refactor? <ins>**Yes**</ins>. Do I really care this much? <ins>**Absolutely not**</ins>. If you still feel like it's a risk to do, let me know, I just _had_ already thought of this so I wanted to put my reasoning out there.

addisonbeck:
> Also, I didn't realise we don't have linting running in CI 😬

We don't run tests in CI either. Both things are noted in [this tech debt item](https://bitwarden.atlassian.net/browse/AC-1998) for auditing the workflows.

jslib/common/src/abstractions/state.service.ts
eliykat: This prompted me to look at the changes closer, and look, you're not wrong. It just didn't seem necessary to me. But, seeing as how you've done it (and explained it), I don't mind going with it.

src/app/accounts/apiKey.component.ts
```diff
@@ -92,12 +95,17 @@ export class ApiKeyComponent {
       this.environmentModal
     );
 
-    childComponent.onSaved.subscribe(() => {
+    childComponent.onSaved.pipe(takeUntil(this.destroyed$)).subscribe(() => {
```
eliykat, line 98: 💭 (non-blocking) Is `takeUntil` the best solution here? You might open and close the Environment dialog multiple times. This starts a new subscription each time, but then would keep all those subscription going until **this** component (ApiKey) is destroyed. Really you only want it going until EnvironmentComponent is destroyed.

eliykat (dismissed):

eliykat:
Removed devops as a reviewer because none of their files have been modified. Assume it was just due to the automatic rebasing.

src/app/accounts/apiKey.component.ts
addisonbeck: Good catch. It's a bit more verbose than I'd like, but I implemented a something on https://github.com/bitwarden/directory-connector/pull/393/commits/baae2b1fbe86872258b08fd006a553b375bdc6e7 for this.

I'm not an rxjs wiz, so open to suggestions if you've got a different idea.

eliykat (changes requested):
I think this would do it, without any need for the `Subject` etc:

```ts
  async settings() {
    const [modalRef, childComponent] = await this.modalService.openViewRef<EnvironmentComponent>(
      EnvironmentComponent,
      this.environmentModal
    );

    childComponent.onSaved.pipe(takeUntil(modalRef.onClosed)).subscribe(() => {
      modalRef.close();
    });
  }
```

modalRef.onClosed will fire whenever the modal is closed, whether by calling `modalRef.close()` in the callback, or by exiting out of the modal some other way.

addisonbeck:
Apologies for taking a while to to get back to this @eliykat, but I took your suggestion as is and did some testing and saw no memory increase when creating and closing a bunch of dialogues. I think we're all set here.

src/app/accounts/apiKey.component.ts
```diff
@@ -17,12 +18,14 @@ import { EnvironmentComponent } from "./environment.component";
...
   templateUrl: "apiKey.component.html",
 })
-export class ApiKeyComponent {
+export class ApiKeyComponent implements OnDestroy {
   @ViewChild("environment", { read: ViewContainerRef, static: true })
   environmentModal: ViewContainerRef;
   @Input() clientId = "";
   @Input() clientSecret = "";
 
+  private modalClosed: Subject<void> = new Subject();
```
eliykat, line 27: You still need to clean up the private subject and `ngOnDestroy` code, which is now unused.

src/app/accounts/apiKey.component.ts
```diff
@@ -87,17 +90,24 @@ export class ApiKeyComponent {
...
 
   async settings() {
-    const [modalRef, childComponent] = await this.modalService.openViewRef(
+    const [modalRef, childComponent] = await this.modalService.openViewRef<EnvironmentComponent>(
       EnvironmentComponent,
       this.environmentModal
     );
 
-    childComponent.onSaved.subscribe(() => {
+    // eslint-disable-next-line rxjs-angular/prefer-takeuntil
```
eliykat, line 98: It didn't occur to me that linting would still expect to see the `onDestroy` pattern specifically. I think we're fine because you shouldn't be able to destroy this component without first closing the dialog, but it would be good to add a comment here explaining why we've disabled the rule.

eliykat (changes requested):

src/app/accounts/apiKey.component.ts
addisonbeck: Good catch. Addressed on https://github.com/bitwarden/directory-connector/pull/393/commits/26c11f30e6be16df9a0626133623ee9c6c3ba0a9

src/app/accounts/apiKey.component.ts
addisonbeck: I thought this was odd to. Removing the `OnDestory` implementation caused another error from the same rule at the component declaration level of the file. I added another disable clause there and put the explanation there. 

This rule seems a little much to me after this experience, but it's in `clients`. The experiences makes me wonder if it would be good practice to just accept some verbosity and do what it says - always including some version of a "close all subscriptions if I get destroyed" even if we don't think there's any way for the subscription to still be open at that time. 

I didn't do that, though. Just rambling. 

Addressed on https://github.com/bitwarden/directory-connector/pull/393/commits/26c11f30e6be16df9a0626133623ee9c6c3ba0a9

eliykat (approved):
I'm also fine with that verbosity, I almost suggested it as an option myself. Happy to re-approve if you want to make that change, or you can get on with it, up to you.

PR #401, opened 2024-01-13
[AC-1743] pt. 5 ⮕ Perform Minor Upgrades

addisonbeck:
# [\[AC-1743\]](https://bitwarden.atlassian.net/browse/AC-1743) pt. 5 ⮕ Perform Minor Upgrades

## Type of change

- [ ] Bug fix
- [ ] New feature development
- [x] Tech debt (refactoring, code cleanup, dependency upgrades, etc)
- [ ] Build/deploy pipeline (DevOps)
- [ ] Other

## 💰 Objective

As part of an initiative to audit dependencies across Directory Connector we should perform minor upgrades on any dependencies that need them.

### 🏦 The Problem

Directory Connector's dependencies are outdated. There are a handful of minor upgrades we can do that should be safe and worry free.

### 🥷 The Solution

The following patch upgrades were made, covering all patch upgrades currently available. 

| Package | Upgraded From | Upgraded To |
| :------ | :-----------: | :---------: |
| `@angular-eslint/eslint-plugin-template` | `17.1.1` | `17.2.0` |
| `@angular-eslint/template-parser` | `17.1.1` | `17.2.0` |
| `@microsoft/microsoft-graph-types` | `2.38.0` | `2.40.0` |
| `core-js` | `3.32.1` | `3.35.0` |
| `css-loader` | `6.8.1` | `6.9.0` |
| `electron-builder` | `24.6.3` | `24.9.1` |
| `eslint` | `8.48.0` | `8.56.0` |
| `eslint-plugin-import` | `2.28.1` | `2.29.1` |
| `html-webpack-plugin` | `5.5.3` | `5.6.0` |
| `jest` | `29.6.4` | `29.7.0` |
| `sass` | `1.66.1` | `1.69.7` |
| `ts-loader` | `9.4.4` | `9.5.1` |
| `webpack` | `5.88.2` | `5.89.0` |
| `webpack-merge` | `5.9.0` | `5.10.0` |

> ℹ️ NOTE: I didn't go through all of these and check for breaking changes, and trusted them to be non-breaking changes following the standard of semantic versioning. To verify that there aren't any breaking changes I am trusting our build automations, tests, QA process, and my own application testing. If we want to do a more thorough job auditing these lets discuss.

## ⏭️ What's next?

Next up in this dependency upgrade effort I'll be:

1. Performing any "safe" major upgrades for outdated packages
1. Upgrading Angular, Webpack, Node, and npm to their latest LTS versions
[...]

PR #403, opened 2024-01-22
[AC-1743] pt. 6 ⮕ Perform "Safe" Major Upgrades

addisonbeck:
# [\[AC-1743\]](https://bitwarden.atlassian.net/browse/AC-1743) pt. 6 ⮕ Perform "Safe" Major Upgrades

> ℹ️ NOTE: This PR is dependent on #401

> ℹ️ NOTE: Most of the lines of code changed on this PR are formatting changes from a `prettier` run using a new version of the package. The true changes here are the package updates and any commits that reference fixing breaking changes.

## Type of change

- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## 🥇 Objective

As part of an initiative to audit dependencies across Directory Connector we should perform any "safe" major upgrades in batch on any dependencies that need them.

### 🏴‍☠️ The Problem

Directory Connector's dependencies are outdated. There are a handful of relatively safe major upgrades we can do that should not pose much risk of a change failure.

### 🤺 The Solution

The following major upgrades were made, covering all dev dependencies currently available. I went through patch notes for each of these, and any breaking changes were documented below and addressed in code if needed.

| Package | Upgraded From | Upgraded To |
| :------ | :-----------: | :---------: |
| commander | 7.20 | 11.1.0 |
| @types/inquirer | 8.2.6 | 9.0.7 |
| @types/ldapjs | 2.2.5 | 3.0.6 |
| @typescript-eslint/eslint-plugin | 5.62.0 | 6.19.1 |
| @typescript-eslint/parser | 5.62.0 | 6.19.1 |
| copy-webpack-plugin | 11.0.0 | 12.0.2 |
| dotenv | 9.0.2 | 16.3.2 |
| eslint-config-prettier | 8.10.0 | 9.1.0 |
| eslint-import-resolver-typescript | 2.7.1 | 3.6.1 |
| html-loader | 3.1.2 | 5.0.0 |
| jest-preset-angular | 13.1.1 | 14.0.0 |
| lint-staged | 12.5.0 | 15.2.0 |
| prettier | 2.8.8 | 3.2.2 |
| sass-loader | 12.6.0 | 14.0.0 |
| typescript-transform-paths | 2.2.4 | 3.4.6 |

#### Breaking Changes

* `commander`
    * `v8.0.0` of commander removed the default export, which had impacts on the way we set up the CLI arguments.
* `inquirer`
    * Somewhere along the way the inquirer call path for a "Separator" was moved to "default.Separator". I couldn't tell when this happened from release notes.
* `prettier`
    * A prettier run generated a lot of changes.
    * `prettier v3.2.3` added a [breaking change](https://github.com/prettier/prettier/releases/tag/3.2.3) that formats `tsconfig.json` as `jsonc`, thus adding trailing commas. `tsconfig.json` doesn't support trailing commas, so this doesn't work. To mitigate we are sticking with `v3.2.2` for right now, which is the same version used in `clients`.
* `eslint`
    * I'm not sure if this is an old issue or not, but eslint started complaining about using `@` symbols in HTML templates and suggested using the HTML code instead. A commit on this PR accepts that feedback.
* `rimraf`
    * This changeset also meant to upgrade `rimraf`, but that caused some unexpected downstream issues with electron and has been excluded from this PR.

## ⏭️ What's next?

Next up in this dependency upgrade effort I'll be:

1. Upgrading Angular, Webpack, Node, Electron, Typescript, and npm to their latest LTS versions. This may happen over multiple PRs depending on breaking changes.

## Testing requirements

There are no specific testing requirements for this. QA will be doing a regression run of Directory Connector at the end of this effort.

## Before you submit


addisonbeck:
This PR is really broken and I'm going to take things a bit slower with these.

PR #417, opened 2024-01-25
Upgrade node (v~18 ⮕ v20.11.0)

addisonbeck:
## Type of change

- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

As part of a dependency audit of the project this PR updates `node` to `v20.11.0`

Brought along for the ride are updates to `npm` and `@types/node` to match the new node version.

There are no breaking changes in this upgrade

## Code changes

* Upgrade versions in `package.json`
* Rebuild `package-lock.json`

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     "@types/ldapjs": "2.2.5",
     "@types/lowdb": "1.0.15",
     "@types/lunr": "2.3.7",
-    "@types/node": "18.17.12",
+    "@types/node": "20.11.0",
     "@types/node-fetch": "2.6.10",
     "@types/node-forge": "1.3.11",
     "@types/proper-lockfile": "4.1.4",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -178,8 +178,8 @@
     "zxcvbn": "4.4.2"
   },
   "engines": {
-    "node": "~18",
-    "npm": "~9"
+    "node": "20.11.0",
+    "npm": "10.2.4"
```
withinfocus, line 182: ❓ What's the best practice here on major-only vs specific versions? I usually just see `~20` for example.

addisonbeck: I hadn't completely made up my mind about this, or read about it. It's written this way in this draft just to experiment a bit. I'm not sure that this PR is going to make it anywhere right now anyway.

Hinton:
FYI, we generally try to stay on the same major node version as electron to decrease any potential side effects that may be introduced (ideally node 18 and 20 behaves identical but you never know) https://www.electronjs.org/docs/latest/tutorial/electron-timelines.

PR #567, opened 2024-08-30
BRE-141 Refactor Release workflow to split deploy/publish steps

urbinaalex17:
## Type of change

- [ ] Bug fix
- [ ] New feature development
- [x] Build/deploy pipeline (BRE)
- [ ] Other

## Objective

Split `directory-connector` GitHub Release from Deploy/Publish.

## Code changes

- **.github/workflows/release.yml:** Remove deploy/publish steps.
- **.github/workflows/publish.yml:** Create new workflow with removed steps from `release.yml`

## Before you submit


.github/workflows/publish.yml
michalchecinski, line 1: I'm wondering if we need a publish workflow at all in `directory-connector` since nothing is technically published/deployed.

urbinaalex17: As discussed in today's standup, even though there are no actual publish/deploy steps, we still need to separate the logic to have a standard process across all the repositories.

vgrassia: @urbinaalex17 The Publish workflow can be deleted.  We don't need it if it's not doing anything.  The Release workflow can stay as you have it with the deployment steps removed for consistency.

.github/workflows/release.yml
```diff
@@ -102,18 +92,3 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           draft: true
```
vgrassia, line 94:
```suggestion
```

vgrassia (changes requested):

.github/workflows/publish.yml
urbinaalex17: Shouldn't we keep the publish workflow for consistency too?

vgrassia (approved):

eliykat (approved):
I don't have a lot of background for this change, but it looks like Github Deployments aren't really relevant to directory-connector because it's a downloadable app only (not hosted anywhere).

PR #637, opened 2024-10-02
[PM-13008] Add ldap integration tests

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-13008

See the parent ticket https://bitwarden.atlassian.net/browse/PM-13007 for additional context.

## 📔 Objective

Add integration tests for `LdapDirectoryService`.

We don't have any integration tests for our directory services, so this was a bit of trial and error, and is open to feedback about the best way to structure these.

The tests use jest because I still needed a test runner and I wanted to isolate `LdapDirectoryService` from the rest of the app and from the Bitwarden server. The integration is purely between our directory service implementation and the external directory service, which to me is the "here be dragons" part of our code at the moment.

This PR uses the OpenLdap docker image that we [already recommend for local development](https://contributing.bitwarden.com/getting-started/business/directory-connector/open-ldap). That's ideal because we can spin it up with seed data on demand. Additional integration tests will probably involve actual third party cloud services or Azure VMs which will require more investigation and setup.

Once this is merged I can remove the OpenLdap configuration from the `server` repository and update the contributing docs. It really belongs with this repo.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


.github/workflows/integration-test.yml
eliykat, line 1: I'm not sure whether this should be separate to the existing `test.yml` file. As we add more integration tests it seems useful to keep it separate, but I'm not sure how this affects code coverage reporting.

As it stands this is mostly a copy/paste of `test.yml`.

We may want to use https://github.com/dorny/paths-filter down the line to selectively run different tests based on which directory service was changed. (The standard path filter only applies to the whole workflow, not steps within it.)

src/models/groupEntry.ts
eliykat, line 1: The `toJSON`/`fromJSON` methods were added as helpers to instantiate the fixture data. It's maybe a bit weird to add this to runtime code, but the `Jsonify` type is based on the `toJSON` return type, so it was the tidiest and least verbose way to add type safety and factory methods to the fixture data.

src/services/ldap-directory.service.integration.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+  let logService: MockProxy<LogService>;
+  let i18nService: MockProxy<I18nService>;
+  let stateService: MockProxy<StateService>;
+
+  let directoryService: LdapDirectoryService;
+
+  beforeEach(() => {
+    logService = mock();
+    i18nService = mock();
+    stateService = mock();
```
eliykat, line 32: Future refactor: it would be good to remove `stateService` from the directory service implementations and pass in their config as arguments instead.

src/services/ldap-directory.service.integration.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+    expect(result).toEqual([groupFixtures, userFixtures]);
+  });
+
+  it("gets users and groups with TLS", async () => {
+    stateService.getDirectory.calledWith(DirectoryType.Ldap).mockResolvedValue(
+      getLdapConfiguration({
+        ssl: true,
+        startTls: true,
+        sslAllowUnauthorized: true, // TODO: this could be more robust if we configured certs correctly
+      }),
```
eliykat, line 57: This is a bit of a half baked SSL config at the moment, I just wanted something in place. Ideally we'd specify a certificate and not use `sslAllowUnauthorized`.

.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,98 @@
...
+  workflow_dispatch:
+  push:
+    branches:
+      - "main"
+    paths:
+      - ".github/workflows/integration-test.yml" # this file
+      - "src/services/ldap-directory.service*" # we only have integration for LDAP testing at the moment
+  pull_request:
+    paths:
+      - ".github/workflows/integration-test.yml" # this file
```
addisonbeck, line 13: Any particular reason not to run these on all PRs and all runs of main? One of the benefits of running tests in CI (to me) is catching things I break that I didn't expect. In this case any changes to directory syncing's core code could also break these tests, which would be useful to know. Using a more generic target for runs would also let you put add this as a step in another job instead of an entire separate action, which would reduce boilerplate.

ftr: I don't think that's good advice in general, and we should have better ways to avoid duplicating boilerplate in actions, but that's an out of scope discussion.

eliykat: The tests are isolated from the rest of directory connector's code - the integration is just between `LdapDirectoryService` and the ldap server itself. All other dependencies are mocked and `LdapDirectoryService` is called directly by jest. So changes to other code would have no affect on this test.

My thinking here is that integration tests are most useful to ensure we're communicating with the third party directory service correctly (each having their own protocol and potentially unique implementation quirks). But provided that we get the expected values out of `IDirectoryService.getEntries`, the rest of the code should be able to be unit tested. Also, the core code is the same for each directory service, so it doesn't need to be tested each time.

What do you think about this choice of testing boundary?

.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,98 @@
...
+          node-version: ${{ steps.retrieve-node-version.outputs.node_version }}
+
+      - name: Install Node dependencies
+        run: npm ci
+
+      # We use isolatedModules: true which disables typechecking in tests
+      # Tests in apps/ are typechecked when their app is built, so we just do it here for libs/
+      # See https://bitwarden.atlassian.net/browse/EC-497
+      - name: Run typechecking
+        run: npm run test:types --coverage
```
addisonbeck, line 70: You probably don't need this, it's not required to run tests and is already done in the main test job.

utils/integration-tests/.env
```diff
@@ -0,0 +1 @@
+COMPOSE_PROJECT_NAME=directory-connector
```
addisonbeck, line 1: What's this for?

eliykat: It gives a nice name to the grouping in docker desktop:

Screenshot 2024-10-07 at 1 31 00 PM

.github/workflows/integration-test.yml
eliykat: Fixed in https://github.com/bitwarden/directory-connector/pull/637/commits/bda1104600db6751a35c2a0ba0521a8053741eb7.

eliykat:
I realised belatedly that the OpenLDAP Docker image is unmaintained since ~3 years ago: https://github.com/osixia/docker-openldap.

I'll look at changing it out for https://hub.docker.com/r/bitnami/openldap, which appears maintained and published by Vmware.

utils/integration-tests/.env
addisonbeck: Does it make more sense to set this in `docker-compose.yml`? Docker's documentation says it will take the top level `name` value [here](https://docs.docker.com/reference/cli/docker/compose/#use--p-to-specify-a-project-name).

Also, if you're planning to use this as the scaffolding location for local development containers it doesn't make a lot of sense to put it in the `integration-tests` folder. I do think it's reasonable to put `docker-compose.yml` (and `.env` if you keep it) in project root.

eliykat: I moved `docker-compose.yml` to the project root as suggested and this automatically set the project name as well.

src/services/ldap-directory.service.integration.spec.ts
eliykat: This is fixed in the latest changes.

eliykat:
@addisonbeck This is ready for review again:
* changed to a maintained openldap server image
* restructured directory structure in line with your suggestions
* configured TLS certs properly 🙌 
* it turns out it wasn't that hard to generate TLS certs in the pipeline, so that has been moved into a script that can be used by both CI and local dev

addisonbeck (approved):

PR #703, opened 2024-12-27
[PM-14360] Import Batching

BTreston:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-14360
## 📔 Objective

This PR implements services that allow the directory connector to batch large import requests into multiple small requests to avoid potential timeouts due to packet size exceeding default size limits for some servers. This implementation splits the requests into batches of 2000 users/groups and POSTs them sequentially.

This change also refactors some of the sync service to be more testable and removes some directory related code that does not belong. Also included are some new unit tests to the new functionality and the sync service.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


BTreston:
@eliykat Whenever you're back and able to give this first pass a look over, I just had a few questions:

- it seems like jest doesnt like ECMAScript Modules (which are used in the googleapis) and plugging it into the existing transformer in the jest config didn't seem to fix it. Not sure how to handle the resulting [error](https://stackoverflow.com/questions/51994111/jest-encountered-an-unexpected-token).
- stripe has a cap on how large a transaction can be (999,999.99 USD) so there's a soft limit to how many users you can have it seems. Have we run into this before?
- in terms of batching groups, do we want to batch based on the number of groups, or the number of users within the groups?

I also still have a few todo's:

- More testing for the sync service (ie. for generateNewHash, buildRequests)
- Refactor the sync call into more testable pieces, potentially move some code that might not belong in this service (looking at some of the directory stuff in this service)

openldap/ldifs/directory-11000.ldif
eliykat, line 1: This directory is used for integration tests. Because your new service is more or less functional - i.e. it's just processing a set of inputs to a set of outputs - I'm not sure we need dedicated integration tests on it.

If you disagree and want to integration test it, I don't think we should maintain a set of 11k users and 11k groups. Ideally you'd specify a lower batchSize and use a more maintanable set of test data (e.g. 250 users with a batchSize of 45). I recall we discussed trying to replicate real world conditions as much as possible, but this is a lot of fixture data to maintain forever. And would be comparatively slow to run in the pipeline.

jslib/common/src/services/batching.service.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+  it("Batches requests for > 2000 users", () => {
+    const mockGroups = groupSimulator(11000);
+    const mockUsers = userSimulator(11000);
+
+    const requests = batchingService.batchRequests(mockGroups, mockUsers, true, true);
+
+    expect(requests.length).toEqual(
+      Math.ceil(mockGroups.length / batchingService.batchSize) +
+        Math.ceil(mockUsers.length / batchingService.batchSize),
+    );
```
eliykat, line 44: I almost always prefer constants in assertions, otherwise you risk rewriting the logic of the SUT.

For example - you know that you've generated 11k users and you have batch sizes of 2k, so you know you're going to have 6 user requests. Assert that your request has a length of 6. (example only, obviously you have groups as well.)

I think it would help with this to specify the `batchSize` in the constructor of the service. That way, the test can specify a fixed batch size and base its assertions on that, without affecting the batch size used by the production code.

jslib/common/src/services/batching.service.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+    };
+
+    groupSimulator = (groupCount: number) => {
+      const simulatedArray: GroupEntry[] = [];
+      for (let i = 0; i < groupCount; i += batchingService.batchSize) {
+        for (let j = 0; j <= batchingService.batchSize; j++) {
+          simulatedArray.push(new GroupEntry());
+        }
+      }
+      return simulatedArray;
```
eliykat, line 31: I don't understand why this needs to know about batchSize - it should just generate the number of mock objects requested. Particularly because this is generating mock input data, which is usually supplied by a directory service which knows nothing about batching.

eliykat (changes requested):
Looking good, here are a few comments but I'll take a closer look once you've finished your draft.

To answer your questions:

> it seems like jest doesnt like ECMAScript Modules (which are used in the googleapis) and plugging it into the existing transformer in the jest config didn't seem to fix it. Not sure how to handle the resulting [error](https://stackoverflow.com/questions/51994111/jest-encountered-an-unexpected-token).

Looking at the [jest documentation](https://jestjs.io/docs/ecmascript-modules), it doesn't have great support for ECMAScript. I think the easiest solution is to decouple the `GsuiteDirectoryService` from `SyncService`, thus avoiding importing the google apis into jest at all. e.g. move `getDirectoryService` to its own class with its own abstraction, and have `SyncService` only inject the abstraction.

This will become a problem again if/when we get a Gsuite testbed set up, but I think that solves it for now, and is a better design.

>    stripe has a cap on how large a transaction can be (999,999.99 USD) so there's a soft limit to how many users you can import and provision seats for at one time. Have we run into this before?

We have not run into this before, but spending $1 million USD in a single provisioning sync seems unlikely. I think we can leave it unhandled.

>    in terms of batching groups, do we want to batch based on the number of groups, or the number of users within the groups?

The goal here is to avoid timeouts at the network edge, so there isn't any hard and fast rule as to how we should split our requests. What do you think?

I can see an argument for splitting based on the number of users in each group - that's the property that scales as you add more users. But my assumption was that these timeouts are actually fairly generous (this request data is very small and this is the first time I've heard of requests timing out), so the simplest approach of batching by _n_ users and _n_ groups would be enough to resolve the issue.

src/services/sync.service.ts
```diff
@@ -133,6 +119,39 @@ export class SyncService {
...
+      "sha256",
+    );
+    if (hashBuff != null) {
+      hash = Utils.fromBufferToB64(hashBuff);
+    }
+
+    return { hash, hashLegacy };
+  }
+
+  async compareToLastHash(hashResult: HashResult): Promise<boolean> {
```
eliykat, line 149: This name tells us it's doing a comparison, but not what that comparison is or how to interpret the returned result.

Suggestion: `isNewHash` or `hashHasChanged` or similar.

src/services/sync.service.ts
```diff
@@ -133,6 +119,39 @@ export class SyncService {
     }
   }
 
+  async generateHash(reqJson: string): Promise<HashResult> {
+    const orgId = await this.stateService.getOrganizationId();
+    if (orgId == null) {
+      throw new Error("Organization not set.");
+    }
+
+    // TODO: Remove hashLegacy once we're sure clients have had time to sync new hashes
```
eliykat, line 128: It's probably too much to do it here, even though it's a simple change, but can you please make a follow-up ticket to remove legacyHash?

jslib/common/src/services/batch-requests.service.ts
eliykat, line 1: The class name should match the filename. e.g. `BatchRequestBuilder` -> `batch-request-builder.ts` or `batch-request.builder.ts`. Or `BatchRequestBuilderService` -> `batch-request-builder.service.ts`. I don't mind as long as it's consistent. We do tend to call things services but not everything needs to be.

eliykat, line 1: Also, please add jsdoc comments to these classes, in particular explaining why we've implemented these different ways of building requests.

src/services/sync.service.spec.ts
```diff
@@ -0,0 +1,188 @@
...
+    const mockRequests = [];
+
+    for (let i = 0; i <= totalUsers; i += batchSize) {
+      mockRequests.push({
+        members: [],
+        groups: [],
+        overwriteExisting: true,
+        largeImport: true,
+      });
+    }
```
eliykat, line 111: I think this would be clearer if you expressly created _n_ requests. Then we don't have to express any batching logic in this test, it's more straightforwardly testing "if given _n_ requests it will send each of them". I think you can just do:

```ts
const mockRequests = new Array(6).fill({
        members: [],
        groups: [],
        overwriteExisting: true,
        largeImport: true,
      });
```

src/services/sync.service.spec.ts
```diff
@@ -0,0 +1,188 @@
...
+
+    batchRequestBuilder.buildRequest.mockReturnValue(mockRequests);
+
+    await syncService.sync(true, false);
+
+    expect(apiService.postPublicImportDirectory).toHaveBeenCalledTimes(6);
```
eliykat, line 117: Suggestion: can you also assert it's been called once with each of the requests?

src/services/sync.service.spec.ts
```diff
@@ -0,0 +1,188 @@
...
+    stateService.getDirectory
+      .calledWith(DirectoryType.Ldap)
+      .mockResolvedValue(getLdapConfiguration());
+
+    stateService.getSync.mockResolvedValue(getSyncConfiguration({ groups: true, users: true }));
+    cryptoFunctionService.hash.mockResolvedValue(new ArrayBuffer(0));
+
+    await syncService.sync(true, false);
+
+    expect(apiService.postPublicImportDirectory).toHaveBeenCalledTimes(0);
```
eliykat, line 130: I assume this works, but I think it's more idiomatic to use:

```suggestion
    expect(apiService.postPublicImportDirectory).not.toHaveBeenCalled();
```

src/services/sync.service.spec.ts
```diff
@@ -0,0 +1,188 @@
...
+    expect(apiService.postPublicImportDirectory).toHaveBeenCalledTimes(6);
+  });
+
+  it("does not post for the same hash", async () => {
+    stateService.getDirectory
+      .calledWith(DirectoryType.Ldap)
+      .mockResolvedValue(getLdapConfiguration());
+
+    stateService.getSync.mockResolvedValue(getSyncConfiguration({ groups: true, users: true }));
+    cryptoFunctionService.hash.mockResolvedValue(new ArrayBuffer(0));
```
eliykat, line 126: I'm missing something here - I can't see that you mock the `stateService.getLastSyncHash` value. How have you mocked/arranged the hash comparison?

src/services/sync.service.spec.ts
```diff
@@ -0,0 +1,188 @@
...
+
+    await syncService.sync(true, false);
+
+    expect(apiService.postPublicImportDirectory).toHaveBeenCalledTimes(0);
+  });
+});
+
+/**
```
eliykat, line 134: From here down looks like a copy/paste of the integration tests. I suggest rather than duplicating it it's moved into a spec helper file for reuse.

src/services/sync.service.ts
```diff
@@ -78,42 +83,23 @@ export class SyncService {
...
         syncConfig.overwriteExisting,
         syncConfig.largeImport,
       );
-      const reqJson = JSON.stringify(req);
 
-      const orgId = await this.stateService.getOrganizationId();
-      if (orgId == null) {
-        throw new Error("Organization not set.");
-      }
+      const reqJson = JSON.stringify(reqs);
```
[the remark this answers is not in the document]
eliykat: Also, this is subtly changing how the has is generated. A single request used to jsonify to an object, now it's jsonifying to an array containing a single object. It's not a huge deal, but for backwards compatibility I think it would be good to "unwrap" a single request from the array before hashing.

src/services/sync.service.ts
```diff
@@ -198,48 +217,28 @@ export class SyncService {
...
-          email: u.email,
-          externalId: u.externalId,
-          deleted: u.deleted || (removeDisabled && u.disabled),
-        };
-      }),
-      overwriteExisting: overwriteExisting,
-      largeImport: largeImport,
-    });
+  ): OrganizationImportRequest[] {
+    if (largeImport && groups.length + users.length > 2000) {
```
eliykat, line 227: This should refer to the `batchSize` constant.

jslib/common/src/services/batch-requests.service.ts
```diff
@@ -0,0 +1,65 @@
+import { OrganizationImportRequest } from "@/jslib/common/src/models/request/organizationImportRequest";
+
+import { GroupEntry } from "@/src/models/groupEntry";
+import { UserEntry } from "@/src/models/userEntry";
+
+import { RequestBuilderAbstratction } from "../abstractions/request-builder.service";
+
+export class BatchRequestBuilder implements RequestBuilderAbstratction {
+  batchSize = 2000;
```
eliykat, line 9: Per previous comment - I suggest this is defined as a constant outside of this class so it can be referenced in `SyncService` as well.

jslib/common/src/services/batch-requests.service.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+  let batchRequestBuilder: BatchRequestBuilder;
+  let singleRequestBuilder: SingleRequestBuilder;
+
+  function userSimulator(userCount: number) {
+    const simulatedArray: UserEntry[] = [];
+    for (let i = 0; i <= userCount; i++) {
+      simulatedArray.push(new UserEntry());
+    }
+    return simulatedArray;
+  }
```
eliykat, line 17: This can be simplified:

```ts
return Array(userCount).fill(new UserEntry());
```

eliykat (changes requested):
This all looks pretty good! Comments are all fairly minor.

src/services/sync.service.spec.ts
BTreston: Since `cryptoFunctionService.hash` returns an empty ArrayBuffer, the result from the generated hash will not satisfy the first condition in line 94 of the sync service. Its focusing on the wrong condition, relying on generating a falsey hash rather than the result from the comparison call being false.

jslib/angular/src/services/jslib-services.module.ts
```diff
@@ -26,6 +26,12 @@ import { StateService } from "@/jslib/common/src/services/state.service";
 import { StateMigrationService } from "@/jslib/common/src/services/stateMigration.service";
 import { TokenService } from "@/jslib/common/src/services/token.service";
 
+import { DirectoryFactoryService } from "@/src/abstractions/directory-factory.service";
+import { StateService as StateServiceExtended } from "@/src/abstractions/state.service";
```
BTreston, line 30: I don't like this naming... open to suggestions. Alternatively we could rename the existing state service since we have 2 (one in jslib and one in src)

eliykat: The src StateService extends the jslib StateService, so you can just use the extended class everywhere and it'll satisfy the type constraints.

However the better answer is that you don't need to add to `jslib-services.module.ts`. Like the rest of the jslib folder, it's essentially deprecated (but I acknowledge this isn't documented anywhere). Register your new dependencies in `src/app/services/services.module.ts`. That file only references the src StateService.

src/services/sync-config-helpers.ts
eliykat, line 1: I think this needs to be more explicitly marked as test data. e.g. `src/utils/test-fixtures.ts`.

src/services/sync.service.spec.ts
```diff
@@ -96,25 +100,27 @@ describe("SyncService", () => {
...
 
     await syncService.sync(true, false);
 
     expect(apiService.postPublicImportDirectory).toHaveBeenCalledTimes(6);
+    expect(apiService.postPublicImportDirectory).toHaveBeenCalledWith({
+      members: [],
+      groups: [],
+      overwriteExisting: true,
+      largeImport: true,
+    });
```
eliykat, line 123: I don't think this should pass - we shouldn't be sending empty request objects to the server.
I expected this would be something like:

```ts
expect(apiService.postPublicImportDirectory).toHaveBeenCalledWith(mockRequests[0]);
expect(apiService.postPublicImportDirectory).toHaveBeenCalledWith(mockRequests[1]);
expect(apiService.postPublicImportDirectory).toHaveBeenCalledWith(mockRequests[2]);
// etc
```

src/services/default-single-request-builder.ts
eliykat, line 1: This is not a big deal, but I could've been more explicit re: naming.

If you had only 1 request builder class, you might do something like:

```ts
DefaultRequestBuilder extends RequestBuilder
```

This is a most classes where you only have a single implementation.

However, here we do have multiple implementations and we already have a way to distinguish them:

```ts
SingleRequestBuilder extends RequestBuilder
BatchRequestBuilder extends RequestBuilder
// there is no "default" implementation, you have to choose one or the other
```

src/services/default-batch-requests-builder.spec.ts
```diff
@@ -1,32 +1,24 @@
 import { GroupEntry } from "@/src/models/groupEntry";
 import { UserEntry } from "@/src/models/userEntry";
 
-import { BatchRequestBuilder } from "./batch-requests.service";
-import { SingleRequestBuilder } from "./single-request.service";
+import { DefaultBatchRequestBuilder } from "./default-batch-request-builder";
+import { DefaultSingleRequestBuilder } from "./default-single-request-builder";
 
 describe("BatchingService", () => {
```
eliykat, line 7: Still using the old name

eliykat (changes requested):
There's also a test failure here that I can't explain. Is it actually trying to use apiService somewhere? https://github.com/bitwarden/directory-connector/runs/36020428086#r0s26

src/services/default-batch-request-builder.ts
eliykat, line 1: You've duplicated this file instead of renaming it (there is now a `default-batch-request-builder.ts` and a `batch-requests-builder.ts`).

src/services/sync.service.spec.ts
```diff
@@ -51,12 +55,12 @@ describe("SyncService", () => {
     stateService.getDirectoryType.mockResolvedValue(DirectoryType.Ldap);
     stateService.getOrganizationId.mockResolvedValue("fakeId");
     directoryFactory.createService.mockReturnValue(
-      new LdapDirectoryService(logService, i18nService, stateService),
+      mockDirectoryService as unknown as IDirectoryService,
     );
```
eliykat, line 59: Here's how to do this with jest without having to write your own mock implementation and mess around with the types not matching:

```ts
const mockDirectoryService = mock<LdapDirectoryService>();
mockDirectoryService.getEntries.mockResolvedValue([groupFixtures, userFixtures]);
directoryFactory.createService.mockReturnValue(ldapDirectoryService);
```

src/services/sync.service.spec.ts
```diff
@@ -51,12 +55,12 @@ describe("SyncService", () => {
...
 
     syncService = new SyncService(
       cryptoFunctionService,
-      apiService,
+      apiService as unknown as ApiService,
```
eliykat, line 63: Why is this type cast necessary?

src/abstractions/request-builder.service.ts
```diff
@@ -0,0 +1,13 @@
+import { OrganizationImportRequest } from "@/jslib/common/src/models/request/organizationImportRequest";
+
+import { GroupEntry } from "@/src/models/groupEntry";
+import { UserEntry } from "@/src/models/userEntry";
+
+export abstract class RequestBuilderAbstratction {
```
eliykat, line 6: `Abstraction` suffix is not required:

```suggestion
export abstract class RequestBuilder {
```

(then your names should be good!)

src/services/sync.service.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+  it("Sync posts multiple request successfully for unique hashes", async () => {
+    stateService.getSync.mockResolvedValue(
+      getSyncConfiguration({ groups: true, users: true, largeImport: true }),
+    );
+    cryptoFunctionService.hash.mockResolvedValue(new ArrayBuffer(1));
+    // This arranges the last hash to be differet from the ArrayBuffer after it is converted to b64
+    stateService.getLastSyncHash.mockResolvedValue("unique hash");
+
+    // @ts-expect-error This is a workaround to make the batchsize smaller to trigger the batching logic since its a const.
+    constants.batchSize = 4;
```
eliykat, line 105: I'd be interested to know if there's a better way of doing this. This feels hacky, but I don't have any better ideas at the moment. 🤔  It's more important that it's readonly at runtime.

eliykat (changes requested):

src/services/sync.service.spec.ts
BTreston: Yeah not a fan of this either, couldn't come up with anything better. Tried doing jest stuff to mock it but none of the myriad solutions offered through searching quite fit the bill.

src/services/sync.service.spec.ts
BTreston: Leftover artifact from trying to fix that one test failure.

eliykat (approved):
Great work, thanks for working through my feedback!

PR #737, opened 2025-04-10
Batch request fix

BTreston:
## 📸 Screenshots
11k users sync before:
https://github.com/user-attachments/assets/9a833bca-a8bf-481e-906f-7bf1a12ff2e6

11k users sync after:
https://github.com/user-attachments/assets/bec6d3ee-4e60-4d16-8088-014bc774ffc8

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/services/batch-requests-builder.spec.ts
+++ b/src/services/batch-requests-builder.spec.ts
@@ -21,26 +21,40 @@ describe("BatchRequestBuilder", () => {
     singleRequestBuilder = new SingleRequestBuilder();
   });
 
+  const defaultOptions = { overwriteExisting: false, removeDisabled: false };
+
   it("BatchRequestBuilder batches requests for > 2000 users", () => {
     const mockGroups = groupSimulator(11000);
     const mockUsers = userSimulator(11000);
 
-    const requests = batchRequestBuilder.buildRequest(mockGroups, mockUsers, true, true);
+    const requests = batchRequestBuilder.buildRequest(mockGroups, mockUsers, defaultOptions);
 
     expect(requests.length).toEqual(12);
+  });
+
+  it("BatchRequestBuilder throws error when overwriteExisting is true", () => {
+    const mockGroups = groupSimulator(11000);
+    const mockUsers = userSimulator(11000);
+    const options = { ...defaultOptions, overwriteExisting: true };
+
+    const r = () => batchRequestBuilder.buildRequest(mockGroups, mockUsers, options);
+
+    expect(r).toThrow(
+      "You cannot use the 'Remove and re-add organization users during the next sync' option with large imports.",
+    );
   });
 
   it("SingleRequestBuilder returns single request for 200 users", () => {
     const mockGroups = groupSimulator(200);
     const mockUsers = userSimulator(200);
 
-    const requests = singleRequestBuilder.buildRequest(mockGroups, mockUsers, true, true);
+    const requests = singleRequestBuilder.buildRequest(mockGroups, mockUsers, defaultOptions);
 
     expect(requests.length).toEqual(1);
   });
 
   it("BatchRequestBuilder retuns an empty array when there are no users or groups", () => {
-    const requests = batchRequestBuilder.buildRequest([], [], true, true);
+    const requests = batchRequestBuilder.buildRequest([], [], defaultOptions);
 
     expect(requests).toEqual([]);
   });
--- a/jslib/common/src/models/request/organizationImportRequest.ts
+++ b/jslib/common/src/models/request/organizationImportRequest.ts
@@ -8,16 +8,12 @@ export class OrganizationImportRequest {
   overwriteExisting = false;
   largeImport = false;
 
-  constructor(
-    model:
-      | {
-          groups: Required<OrganizationImportGroupRequest>[];
-          users: Required<OrganizationImportMemberRequest>[];
-          overwriteExisting: boolean;
-          largeImport: boolean;
-        }
-      | ImportDirectoryRequest,
-  ) {
+  constructor(model: {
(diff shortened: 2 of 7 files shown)


eliykat (changes requested):
Can you please add some unit tests for the following:
- SingleRequestBuilder - make sure that the options (removeDisabled and overwriteExisting) are set properly on the request object based on inputs
- BatchRequestBuilder - make sure that removeDisabled is set properly on the request object based on inputs (you've already covered overwriteExisting which is good)

Can you also consider what integration tests we can add for these? I didn't think it was worth it previously because these are directory-agnostic and unit tests should be enough. But our recent experience shows that a few happy path integration tests would also be a good safeguard.

I want to set up full end-to-end integration tests against a real server, but for now we can just work within the openldap test suite we already have.

src/services/batch-requests-builder.spec.ts
```diff
@@ -44,13 +46,22 @@ describe("BatchRequestBuilder", () => {
...
-    expect(requests.length).toEqual(1);
+  it("BatchRequestBuilder returns requests with deleted users when removeDisabled is true", () => {
+    const mockGroups = groupSimulator(11000);
+    const disabledUser = new UserEntry();
+    const mockUsers = userSimulator(11000);
+    const options = new RequestBuilderOptions({ ...defaultOptions, removeDisabled: true });
+
+    disabledUser.disabled = true;
+    mockUsers[0] = disabledUser;
+    mockUsers.push(disabledUser);
```
eliykat, line 57: I think it's better to make specific `UserEntry`s with unique identifiers (i.e. email) so that you can check that those specific users are disabled. Same in the single request builder tests.

src/services/single-request-builder.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+describe("SingleRequestBuilder", () => {
+  let singleRequestBuilder: SingleRequestBuilder;
+
+  function userSimulator(userCount: number) {
+    return Array(userCount).fill(new UserEntry());
+  }
+
+  function groupSimulator(groupCount: number) {
+    return Array(groupCount).fill(new GroupEntry());
+  }
```
eliykat, line 17: Please extract these to a common file to be reused between the two test suites.
It would also be good to at least populate the email/group name so they approximate the real thing a bit more closely. e.g. `GetUniqueString() + "@example.com"` (see `jslib/common/spec/utils.ts`).

src/services/single-request-builder.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  beforeEach(async () => {
+    singleRequestBuilder = new SingleRequestBuilder();
+  });
+
+  const defaultOptions = new RequestBuilderOptions({
+    overwriteExisting: false,
+    removeDisabled: false,
+  });
```
eliykat, line 26: This object risks being mutated between tests. You should re-instantiate it each time or `Object.freeze` it.

src/services/ldap-directory.service.integration.spec.ts
```diff
@@ -75,6 +115,60 @@ describe("ldapDirectoryService", () => {
...
+          overwriteExisting: false,
+        }),
+      );
+
+      cryptoFunctionService.hash.mockResolvedValue(new ArrayBuffer(1));
+      // This arranges the last hash to be differet from the ArrayBuffer after it is converted to b64
+      stateService.getLastSyncHash.mockResolvedValue("unique hash");
+
+      // @ts-expect-error This is a workaround to make the batchsize smaller to trigger the batching logic since its a const.
+      constants.batchSize = 4;
```
eliykat, line 162: Can you save the initial batchsize to a local variable, then re-assign it back at the end of the test? To basically reset the state.

src/services/ldap-directory.service.integration.spec.ts
eliykat, line 1: I forgot these tests are testing the directory service specifically, whereas our new tests are testing syncService. Please move them to a new spec file (`sync.service.integration.spect.ts`) but still use the same ldap container setup.

eliykat (changes requested):

src/services/batch-requests-builder.spec.ts
```diff
@@ -48,20 +41,34 @@ describe("BatchRequestBuilder", () => {
...
+    expect(requests[0].members.find((m) => m.deleted && m.email === disabledUserEmail1)).toEqual(
+      expect.objectContaining({ email: disabledUserEmail1, deleted: true }),
+    );
     expect(requests[1].members.find((m) => m.deleted)).toBeUndefined();
     expect(requests[3].members.find((m) => m.deleted)).toBeUndefined();
     expect(requests[4].members.find((m) => m.deleted)).toBeUndefined();
-    expect(requests[5].members.find((m) => m.deleted)).toBeTruthy();
+    expect(requests[5].members.find((m) => m.deleted && m.email === disabledUserEmail2)).toEqual(
+      expect.objectContaining({ email: disabledUserEmail2, deleted: true }),
+    );
```
eliykat, line 71: The `find` callback is repeating the `toEqual` logic. What you want is `toContainEqual`:

```ts
    expect(requests[5].members).toContainEqual({ email: disabledUserEmail2, deleted: true });
```

(same goes for the previous assertion a few lines above)

src/services/ldap-directory.service.integration.spec.ts
eliykat, line 1: This spec file still has changes that can probably be reverted now that the new tests are in a separate file.

eliykat (changes requested):
Some very minor cleanup, then good to go 🎉

eliykat (approved):
Great work!

PR #797, opened 2025-06-18
[PM-21187] Rename Azure AD to Entra ID

jrmccannon:
## 🎟️ Tracking
[PM-21187](https://bitwarden.atlassian.net/browse/PM-21187)

## 📔 Objective
This renames Azure AD to Entra ID. It includes file names and commands. Existing configs should continue to work, but will be assigned to entra config behind the scenes.

## 📸 Screenshots
image

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-21187]: https://bitwarden.atlassian.net/browse/PM-21187?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/tabs/settings.component.ts
+++ b/src/app/tabs/settings.component.ts
@@ -59,8 +59,7 @@ export class SettingsComponent implements OnInit, OnDestroy {
       this.gsuite;
     this.entra =
       (await this.stateService.getDirectory<EntraIdConfiguration>(DirectoryType.EntraID)) ||
-      this.entra ||
-      this.azure;
+      this.entra;
     this.okta =
       (await this.stateService.getDirectory<OktaConfiguration>(DirectoryType.Okta)) || this.okta;
     this.oneLogin =
--- a/src/app/tabs/settings.component.ts
+++ b/src/app/tabs/settings.component.ts
@@ -42,7 +42,7 @@ export class SettingsComponent implements OnInit, OnDestroy {
     this.directoryOptions = [
       { name: this.i18nService.t("select"), value: null },
       { name: "Active Directory / LDAP", value: DirectoryType.Ldap },
-      { name: "Azure Active Directory", value: DirectoryType.AzureActiveDirectory },
+      { name: "EntraID", value: DirectoryType.AzureActiveDirectory },
       { name: "G Suite (Google)", value: DirectoryType.GSuite },
       { name: "Okta", value: DirectoryType.Okta },
       { name: "OneLogin", value: DirectoryType.OneLogin },
--- a/src/app/tabs/settings.component.ts
+++ b/src/app/tabs/settings.component.ts
@@ -22,7 +22,6 @@ export class SettingsComponent implements OnInit, OnDestroy {
   directoryType = DirectoryType;
   ldap = new LdapConfiguration();
   gsuite = new GSuiteConfiguration();
-  azure = new EntraIdConfiguration();
   entra = new EntraIdConfiguration();
   okta = new OktaConfiguration();
   oneLogin = new OneLoginConfiguration();
--- a/src/app/tabs/settings.component.ts
+++ b/src/app/tabs/settings.component.ts
@@ -5,7 +5,7 @@ import { LogService } from "@/jslib/common/src/abstractions/log.service";
 
 import { StateService } from "../../abstractions/state.service";
 import { DirectoryType } from "../../enums/directoryType";
-import { AzureConfiguration } from "../../models/azureConfiguration";
+import { EntraIdConfiguration } from "../../models/entraIdConfiguration";
 import { GSuiteConfiguration } from "../../models/gsuiteConfiguration";
 import { LdapConfiguration } from "../../models/ldapConfiguration";
 import { OktaConfiguration } from "../../models/oktaConfiguration";
@@ -22,7 +22,7 @@ export class SettingsComponent implements OnInit, OnDestroy {
   directoryType = DirectoryType;
   ldap = new LdapConfiguration();
   gsuite = new GSuiteConfiguration();
-  azure = new AzureConfiguration();
+  azure = new EntraIdConfiguration();
   okta = new OktaConfiguration();
   oneLogin = new OneLoginConfiguration();
   sync = new SyncConfiguration();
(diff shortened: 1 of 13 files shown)


src/services/state.service.ts
```diff
@@ -106,11 +107,11 @@ export class StateService
           await this.setLdapConfiguration(ldapConfig);
           break;
         }
-        case DirectoryType.AzureActiveDirectory: {
-          const azureConfig = config as AzureConfiguration;
-          await this.setAzureKey(azureConfig.key);
+        case DirectoryType.EntraID: {
+          const azureConfig = config as EntraIdConfiguration;
```
BTreston: should probably be renamed:
```suggestion
          const entraIdConfig = config as EntraIdConfiguration;
```

BTreston (changes requested):
1 comment, otherwise looks good

src/services/state.service.ts
jrmccannon: One always sneaks through

src/app/tabs/settings.component.ts
```diff
@@ -56,10 +57,10 @@ export class SettingsComponent implements OnInit, OnDestroy {
...
       (await this.stateService.getDirectory<GSuiteConfiguration>(DirectoryType.GSuite)) ||
       this.gsuite;
-    this.azure =
-      (await this.stateService.getDirectory<AzureConfiguration>(
-        DirectoryType.AzureActiveDirectory,
-      )) || this.azure;
+    this.entra =
+      (await this.stateService.getDirectory<EntraIdConfiguration>(DirectoryType.EntraID)) ||
+      this.entra ||
+      this.azure;
```
eliykat: I don't think you need to keep the `this.azure` property. It's just a default in-memory config that the app uses as a fallback, which is now `this.entra`. It's not a reference to persistent state that we need to be backwards compatible with.

src/models/account.ts
```diff
@@ -29,7 +29,7 @@ export class ClientKeys {
 export class DirectoryConfigurations {
   ldap: LdapConfiguration;
   gsuite: GSuiteConfiguration;
-  azure: AzureConfiguration;
+  entra: EntraIdConfiguration;
```
eliykat, line 32: I believe that _this_ is the class that is serialized and saved in (and read back from) `data.json`. By changing `azure` to `entra` here, it won't be able to find the existing value.

I suggest this:

```suggestion
  // Azure Active Directory was renamed to Entra ID, but we've kept the old account property name
  // to be backwards compatible with existing configurations.
  azure: EntraIdConfiguration;
```

The state migration service is meant for performing migrations in cases like this, but the copy in this codebase is old and unmaintained so I don't think it's worth trying to use here.

src/commands/config.command.ts
```diff
@@ -55,7 +56,8 @@ export class ConfigCommand {
           await this.setGSuiteKey(value);
           break;
         case "azure.key":
-          await this.setAzureKey(value);
+        case "entra.key":
+          await this.setEntraIdKey(value);
```
eliykat, line 61: I agree with this 👍 but please add a comment explaining the rename / backwards compatibility.

src/commands/config.command.ts
```diff
@@ -127,10 +129,10 @@ export class ConfigCommand {
...
       (await this.stateService.getDirectory<GSuiteConfiguration>(DirectoryType.GSuite)) ||
       this.gsuite;
-    this.azure =
-      (await this.stateService.getDirectory<AzureConfiguration>(
-        DirectoryType.AzureActiveDirectory,
-      )) || this.azure;
+    this.entra =
+      (await this.stateService.getDirectory<EntraIdConfiguration>(DirectoryType.EntraID)) ||
+      this.entra ||
+      this.azure;
```
eliykat: Similar to the settings component above, not sure keeping `azure` here is doing anything.

src/services/state.service.ts
```diff
@@ -22,6 +22,7 @@ const SecureStorageKeys = {
   ldap: "ldapPassword",
   gsuite: "gsuitePrivateKey",
   azure: "azureKey",
```
eliykat, line 26: Please add comment re: backwards compatibility

src/services/stateMigration.service.ts
```diff
@@ -104,13 +105,13 @@ export class StateMigrationService extends BaseStateMigrationService {
...
 
-    // Initilize typed objects from key/value pairs in storage to either be saved temporarily until an account is authed or applied to the active account
+    // Initialize typed objects from key/value pairs in storage to either be saved temporarily until an account is authed or applied to the active account
     const getDirectoryConfig = async <T>(type: DirectoryType) =>
       await this.get<T>(SecureStorageKeys.directoryConfigPrefix + type);
     const directoryConfigs: DirectoryConfigurations = {
       ldap: await getDirectoryConfig<LdapConfiguration>(DirectoryType.Ldap),
       gsuite: await getDirectoryConfig<GSuiteConfiguration>(DirectoryType.GSuite),
-      azure: await getDirectoryConfig<AzureConfiguration>(DirectoryType.AzureActiveDirectory),
+      entra: await getDirectoryConfig<EntraIdConfiguration>(DirectoryType.EntraID),
```
eliykat: This is real old point-in-time migration code which is probably never being run anymore. We need to update the type and enum names (necessarily) but don't update anything else - i.e. leave the key as `azure`.

eliykat (changes requested):

eliykat:
When QA test this, please get ask them to test the upgrade path to ensure that the Azure/Entra config and key aren't lost when upgrading between versions.

src/app/tabs/settings.component.ts
```diff
@@ -22,13 +22,14 @@ export class SettingsComponent implements OnInit, OnDestroy {
   directoryType = DirectoryType;
   ldap = new LdapConfiguration();
   gsuite = new GSuiteConfiguration();
-  azure = new AzureConfiguration();
+  azure = new EntraIdConfiguration();
```
eliykat: 🧹 you can also delete the property definition

src/commands/config.command.ts
```diff
@@ -20,7 +20,8 @@ export class ConfigCommand {
   private directory: DirectoryType;
   private ldap = new LdapConfiguration();
   private gsuite = new GSuiteConfiguration();
-  private azure = new AzureConfiguration();
+  private azure = new EntraIdConfiguration();
```
eliykat: And here

PR #815, opened 2025-07-02
Add top level permissions key to Github workflows

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-21333
## 📔 Objective

Add a top level `permissions` key to all Github workflows per our latest best practice.

The following workflows already had this key added and the existing permissions looked reasonable:
- build (contents: read)
- enforce-labels (contents: read, pull-requests: read)
- release (contents: read)
- scan (none)

I have added or adjusted permissions for the following. However, this is based on my best guess based on what the workflows do, so a second set of eyes from BRE would be appreciated. Further in-line comments below.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,16 +9,15 @@ on:
       - "hotfix-rc"
   pull_request:
 
+permissions:
+  contents: read
+
 jobs:
 
   testing:
     name: Run tests
     if: ${{ startsWith(github.head_ref, 'version_bump_') == false }}
     runs-on: ubuntu-24.04
-    permissions:
-      checks: write
-      contents: read
-      pull-requests: write
 
     steps:
       - name: Check out repo
--- a/.github/workflows/version-bump.yml
+++ b/.github/workflows/version-bump.yml
@@ -8,6 +8,9 @@ on:
         required: false
         type: string
 
+permissions:
+  contents: write
+
 jobs:
   bump_version:
     name: Bump Version
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -16,17 +16,13 @@ on:
       - "src/services/ldap-directory.service*" # we only have integration for LDAP testing at the moment
       - "./openldap/**/*" # any change to test fixtures
       - "./docker-compose.yml" # any change to Docker configuration
-
+permissions:
+  contents: read
 jobs:
-
   testing:
     name: Run tests
     if: ${{ startsWith(github.head_ref, 'version_bump_') == false }}
     runs-on: ubuntu-22.04
-    permissions:
-      checks: write
-      contents: read
-      pull-requests: write
 
     steps:
       - name: Check out repo


.github/workflows/version-bump.yml
```diff
@@ -8,6 +8,9 @@ on:
         required: false
         type: string
 
+permissions:
+  contents: write
+
```
eliykat: Looking at the underlying action, it bumps the version and pushes its commit directly to `main`, so I assume this requires `contents: write`. I'm not sure I can test this though.

As a side note, it would be helpful if our Github Actions listed what permissions they required in their README.

.github/workflows/test.yml
```diff
@@ -9,16 +9,16 @@ on:
...
 jobs:
 
   testing:
     name: Run tests
     if: ${{ startsWith(github.head_ref, 'version_bump_') == false }}
     runs-on: ubuntu-24.04
-    permissions:
-      checks: write
-      contents: read
-      pull-requests: write
```
eliykat, line 21: - moved existing permissions to top level to comply with linter (or would it better to disable permissions at the top level and keep the permissions specific to the job?)
- I don't know why it would need `pull-requests: write` so I have removed it
- confirmed that `checks: write` is required by the test reporter. Its [documentation](https://github.com/dorny/test-reporter?tab=readme-ov-file#example) says it also requires `actions: read`, however we've never had that permission (here or in `clients`) and it hasn't caused any harm, so I haven't added it here.

fntyler: > moved existing permissions to top level to comply with linter (or would it better to disable permissions at the top level and keep the permissions specific to the job?)

I'm inclined to think that either would suffice. I lean towards top-level defined permissions, in this case, because there is only one job. In the case of 2 or more jobs, disabling permissions at the top-level and defining [job-specific permissions](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idpermissions) for each seems to make more sense.

.github/workflows/version-bump.yml
pixman20: Correct, I have an [overlapping PR](https://github.com/bitwarden/directory-connector/pull/796/files#diff-e8ca070981ccdc3f369851c960bec6882143a0c7afe78acd7b05e0c2b0068d72) that's waiting on some other things that has a similar change, except I would recommend putting `contents: write` permissions at the job level so that it doesn't unnecessarily grant future jobs those permissions.

eliykat: I've moved the permissions to the job level as recommended, I also brought across the `id-token: write` permission from your PR on the assumption it's needed today (without your other changes).

I didn't see your PR but I think this is the only place we've overlapped, so I assume this PR is still useful to cover off the `permissions` linting requirement.

r-tome (approved):

PR #816, opened 2025-07-08
Split build and release workflows for gui and cli

eliykat:
## 🎟️ Tracking

N/A but I'll create a ticket if we accept this proposal
## 📔 Objective

Split the build and release workflows by client, so that we can build and release CLI and GUI clients separately.

My primary reason here is maintainability - the build workflow is currently 651 LOC with 6 different build jobs for all the combinations of `platform x app`, which is hard to keep track of in a yaml file. This change also matches how our monorepo workflows are structured; avoids mixing app-specific setup steps; and would allow for separate releases if required (e.g. if we had a CLI/GUI specific bug we wanted to fix).

The downside is some duplication in setup and completion jobs in each workflow, but I think this is minimal. There may also be some process overhead when we do a release which we can discuss as a team.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


eliykat:
This would have some downstream consequences we don't want to deal with right now - we have more important things to address in this repo so we may come back to it later.

PR #817, opened 2025-07-08
[PM-21956] Replace vercel/pkg and pkg-fetch with yao-pkg fork

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/jira/dashboards/10012

## 📔 Objective

Replace the archived `pkg` and `pkg-fetch` dependencies with the `yao-pkg` fork that we're using in `clients`.
- `pkg` is used to bundle our CLI app into a single application with its own node runtime
- `pkg-fetch` provides a patched node runtime for this purpose

This also has the effect of upgrading the packaged node runtime from 18 -> 20. We can't upgrade any further until `clients` does as we are still evaluating the node 22 `pkg-fetch` artifact and other alternatives.

Other changes:
- delete the CLOC job because all it does is output a count of the lines of code, which we do not use
- use the node version in `.nvmrc` in the workflow (build environment only) - consistent with `clients`
- align `.nvmrc` and `package.json` node versions

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


.github/workflows/build.yml
```diff
@@ -44,15 +29,23 @@ jobs:
...
     runs-on: ubuntu-24.04
     needs: setup
     env:
       _PACKAGE_VERSION: ${{ needs.setup.outputs.package_version }}
-      _PKG_FETCH_NODE_VERSION: 18.5.0
-      _PKG_FETCH_VERSION: 3.4
+      _NODE_VERSION: ${{ needs.setup.outputs.node_version }}
+      # PKG_FETCH versions must NOT be upgraded until clients has upgraded as we must verify the yao-pkg node builds ourselves
+      _LINUX_PKG_FETCH_VERSION: 3.5 # pkg-fetch release
+      _LINUX_PKG_FETCH_NODE_VERSION: 20.11.1 # node version in the pkg-fetch release
```
Hinton: FYI this most likely does nothing.

Inspecting these in a hex editor shows they use node 20.18.0.

Image

PKG will attempt to use the host version range to satisfy the version. In our case 20.18.0 is the last published version supported by the transient version of yao-fetch.

This is the same as client though so should be fine, but we should remove any reference in the workflow to avoid confusion.

eliykat:
@Hinton Changes since your last review:

- I bumped `yao-pkg` to the latest version audited by @mandreko-bitwarden
- I tried to update to node 22 but that broke keytar, so reverted to node 20 for now (which is still on LTS)
- I removed the explicit `pkg-fetch` steps from the workflows as this file was not being used. We're relying on `pkg` to fetch its own binary instead (which in turn comes from pkg-fetch anyway, but we don't have to manage it)
- I removed the Resource Hacker step from the Windows workflow as this was only modifying the `pkg-fetch` binary that we weren't using anyway. I verified this by checking the version info with Resource Hacker on our latest release:

Screenshot 2025-07-11 131145

However I can create a follow-up ticket to actually fix this.

package.json
```diff
@@ -175,7 +175,7 @@
     "zone.js": "0.14.10"
   },
   "engines": {
-    "node": "~22.13.0",
+    "node": "~20",
```
JimmyVo16, line 178: Non-blocker, just curious why we don't lock down the Node version like we do with dependencies?

JimmyVo16 (dismissed):

package.json
eliykat: This doesn't really have the effect of pinning node. It just kinda yells at you if you're not using the correct version in your dev environment.

The production builds work as follows:
1. Electron GUI app uses the node version packaged with Electron (not affected by this PR)
2. Node CLI app uses the node version packaged by `pkg` (this PR). This will match the major version installed by the build workflow, which in turn is set by `.nvmrc`. `pkg` doesn't provide any good way of specifically pinning the node version, but also `pkg` itself is pinned, so that effectively pins it to the latest node version in that version of `pkg`. (all very roundabout but provides some certainty)

TL;DR - `package.json` `engines` isn't very important.

eliykat:
Putting a hold on this until @Hinton reviews and approves.

Hinton:
@eliykat can we revert to the previous pkg version? I believe we only audited the latest v22 not v20 for it.

eliykat:
Downgraded to 5.16.1 which is the current version used by `clients`.

Hinton (approved):

JimmyVo16 (approved):

PR #838, opened 2025-07-23
[PM-24054] Update GUI build jobs to use nvmrc version

eliykat:
## 📔 Objective

In #817 I changed the build workflow to use the node version in `.nvmrc`. However, I only updated the CLI build jobs, not the GUI build jobs, which are still using node 18. Electron ships with its own version of node so it doesn’t affect the node runtime in the artifact, but the build environment should be using a consistent version of node in all jobs.

This PR updates the GUI build jobs to also use the correct version of node.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -267,6 +267,7 @@ jobs:
     env:
       NODE_OPTIONS: --max_old_space_size=4096
       _PACKAGE_VERSION: ${{ needs.setup.outputs.package_version }}
+      _NODE_VERSION: ${{ needs.setup.outputs.node_version }}
       HUSKY: 0
     steps:
       - name: Checkout repo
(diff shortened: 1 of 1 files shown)


.github/workflows/integration-test.yml
```diff
@@ -10,12 +10,14 @@ on:
       - "src/services/ldap-directory.service*" # we only have integration for LDAP testing at the moment
       - "./openldap/**/*" # any change to test fixtures
       - "./docker-compose.yml" # any change to Docker configuration
+      - "./package.json" # dependencies
```
eliykat, line 13: Opportunistic change while I'm here. I noticed the integration tests weren't running when dependencies were upgraded, which is a good use case for them.

PR #879, opened 2025-09-17
[PM-25903] [PM-25590] BWDC doesn't sync new users unless cache is cleared

sso-bitwarden:
## 📔 Objective

The purpose of the cache is not to make API call if there is no new changes to the directory data. 

Bug: The directory connector never sync data after the first initial sync. This is because there is a condition the prevented that.

The fix: Remove that condition because the hashLegacy is never used anyway. 

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/src/services/sync.service.ts
+++ b/src/services/sync.service.ts
@@ -142,7 +142,7 @@ export class SyncService {
   async isNewHash(hashResult: HashResult): Promise<boolean> {
     const lastHash = await this.stateService.getLastSyncHash();
 
-    return lastHash == null || (hashResult.hash !== lastHash && hashResult.hashLegacy !== lastHash);
+    return lastHash == null || hashResult.hash !== lastHash;
   }
 
   private removeDuplicateUsers(users: UserEntry[]) {


JimmyVo16:
@sso-bitwarden Can you fill out the Objective, including details of the problem and solution? Also, add any screenshots of test artifacts (if there are any) or anything else you feel is relevant.

src/services/sync.service.ts
```diff
@@ -142,7 +142,7 @@ export class SyncService {
   async isNewHash(hashResult: HashResult): Promise<boolean> {
     const lastHash = await this.stateService.getLastSyncHash();
 
-    return lastHash == null || (hashResult.hash !== lastHash && hashResult.hashLegacy !== lastHash);
+    return lastHash == null || hashResult.hash !== lastHash;
```
eliykat, line 145: As discussed, it doesn't seem like this will fix the issue: `hashResult.hashLegacy !== lastHash` will always evaluate to `true` (not false) because nobody should be using the legacy hash anymore; but that means it could never block a sync here.

PR #1012, opened 2026-02-24
[PM-12436] replace node keytar 

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-12436

## 📔 Objective

Rid of node keytar from Directory Connector


BTreston:
@JaredScar can you address the conflicts at your earliest convenience? I want to try and test this all together before moving on to updating all our deps as the next step. The migration for this should come after the new state migration (aka, this pr should be state version 6)

jslib/electron/src/services/electronRendererSecureStorage.service.ts
```diff
@@ -1,43 +1,53 @@
-import { ipcRenderer } from "electron";
 
 import { StorageService } from "@/jslib/common/src/abstractions/storage.service";
 import { StorageOptions } from "@/jslib/common/src/models/domain/storageOptions";
 
+import { passwords } from "dc-native";
+
+const APPLICATION_NAME = "Bitwarden Directory Connector";
```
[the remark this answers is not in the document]
JaredScar: Do we want to act on this? @BTreston

.claude/settings.json
SaintPatrck, line 1: ❓ Was this intentionally included? It doesn't seem related to the other changes, unless I'm overlooking one of our plugin references somewhere.

jslib/electron/src/services/electronRendererSecureStorage.service.ts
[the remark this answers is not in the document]
BTreston: yes, `state.model.ts` is probably a good home for this at the end of the file.

.claude/settings.json
theMickster: Looks like maybe a wonky merge from main? 🤔

JaredScar: I think this was a wonky merge from main... My bad, will fix

theMickster (changes requested):
❌ Please undo all changes to the `.claude/settings.json` file

theMickster (dismissed):
Approved from an AI-SME perspective. Thanks for fixing the wonky merge

BTreston (commented):
## Code Review Summary -- Test Failure Analysis

**PR**: [PM-12436] replace node keytar
**Failing Check**: `Run tests` (ubuntu-24.04, `test.yml` workflow)
**Passing Checks**: Linter, typechecking, SonarCloud, Codecov

---

### Root Cause of Test Failure

The test suite `libs/services/state-service/stateMigration.service.spec.ts` crashes **before any test runs** with:

```
Could not find dc-native binary for linux-x64. Run 'npm run build:native' to compile it.

  at loadFirstAvailable (native/index.js:16:9)
  at Object.<anonymous> (native/index.js:49:25)
  at Object.<anonymous> (libs/services/state-service/stateMigration.service.ts:8:1)
  at Object.<anonymous> (libs/services/state-service/stateMigration.service.spec.ts:5:1)
```

**The import chain causing the failure:**
1. `stateMigration.service.spec.ts` (line 5) imports `StateMigrationService`
2. `stateMigration.service.ts` (line 8) has a **top-level** `import { passwords } from "dc-native"`
3. `native/index.js` eagerly tries to load a `.node` binary file on module evaluation
4. The CI test workflow (`test.yml`) runs on `ubuntu-24.04` and does **not** install Rust or compile the native module
5. The binary `dc_native.linux-x64-gnu.node` does not exist, so `loadFirstAvailable()` throws

The `passwords` object is only used in one place: `migrateStateFrom5To6()` (line 276). However, because it is imported at the module top level, it forces the native binary to load whenever **any** code imports `StateMigrationService`, including tests.

---

### Recommended Fix

**Two changes are needed:**

#### 1. Add a Jest mock for `dc-native` (required to unblock tests)

Add to `jest.config.cjs` in the existing `moduleNameMapper` section, or create a manual mock at `__mocks__/dc-native.js`:

**Option A -- `jest.config.cjs` (simpler):**
Add a `moduleNameMapper` entry that maps `dc-native` to a mock. Since `pathsToModuleNameMapper` generates the mapper object, you can spread it and add the extra entry:

```js
moduleNameMapper: {
  ...pathsToModuleNameMapper(compilerOptions.paths, { prefix: "<rootDir>/" }),
  "^dc-native$": "<rootDir>/__mocks__/dc-native.js",
},
```

Then create `__mocks__/dc-native.js`:

```js
module.exports = {
  passwords: {
    getPassword: jest.fn().mockResolvedValue(""),
    setPassword: jest.fn().mockResolvedValue(undefined),
    deletePassword: jest.fn().mockResolvedValue(undefined),
    isAvailable: jest.fn().mockResolvedValue(true),
    migrateKeytarPassword: jest.fn().mockResolvedValue(false),
    PASSWORD_NOT_FOUND: "Password not found.",
  },
};
```

**Option B -- `jest.mock()` in the spec file (more targeted but less reusable):**
Add a `jest.mock("dc-native", ...)` call at the top of `stateMigration.service.spec.ts`.

#### 2. Consider lazy-loading `dc-native` in production code (recommended but not blocking)

Move the `import { passwords } from "dc-native"` from the top of `stateMigration.service.ts` into the `migrateStateFrom5To6()` method as a dynamic require:

```ts
protected async migrateStateFrom5To6(): Promise<void> {
  const { passwords } = require("dc-native");
  // ... rest of method
}
```

This avoids loading the native binary at module scope, which:
- Makes the module inherently testable without mocks for tests that don't exercise v5-to-v6 migration
- Avoids loading the native binary on platforms/contexts where this migration is a no-op (macOS, Linux)
- Follows the principle of loading heavy native dependencies only when actually needed

---

### Why Only One "Run tests" Check Fails

The CI shows two `Run tests` entries: one from the **Testing** workflow (`test.yml`, ubuntu-only, no Rust) and one from a separate workflow. The `test.yml` workflow only does `npm ci` and `npm run test` -- it does not install Rust or compile native modules, which is why the native binary is missing.

---

### No Other New Issues Found

The latest commit (`4c02902`) addressed several prior review findings (linter fixes, electron hot-reload fix, state v6 test updates, removal of unused `keytarSecureStorageService`). The remaining changes look correct. The test failure is the only blocking issue.

libs/services/state-service/stateMigration.service.ts
```diff
@@ -244,6 +252,48 @@ export class StateMigrationService {
...
+        `${clientId}_${SecureStorageKeys.oneLogin}`,
+      );
+    }
+
+    await Promise.all(
+      credentialKeys.map((key) =>
+        passwords.migrateKeytarPassword(SECURE_STORAGE_SERVICE_NAME, key),
+      ),
+    );
+
```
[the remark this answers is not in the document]
BTreston: @JaredScar please see if this is a real concern (I recently made a fix to migrations upstream, so this may be accurate)

mandreko-bitwarden (dismissed):

gitclonebrian (dismissed):
🚀

libs/services/nativeSecureStorage.service.ts
```diff
@@ -0,0 +1,41 @@
...
+        throw e;
+      });
+  }
+
+  async has(key: StorageKey | SecureStorageKey): Promise<boolean> {
+    return (await this.get(key)) != null;
+  }
+
+  async save(key: StorageKey | SecureStorageKey, obj: any): Promise<any> {
+    if (!obj) {
```
eliykat: Falsey values can be valid to save (e.g. `false`, `0`). I think this should be a more explicit null/undefined check here. (also check any other storage service)

package.json
eliykat, line 1: If we're updating to node 22, yao-pkg also needs to be updated: see [my comment here](https://github.com/bitwarden/directory-connector/pull/953#pullrequestreview-3590073647) for more context

src-gui/main.ts
```diff
@@ -48,8 +49,11 @@ export class Main {
     const watch = args.some((val) => val === "--watch");
 
     if (watch) {
-      // eslint-disable-next-line
-      require("electron-reload")(__dirname, {});
+      fs.watch(__dirname, { recursive: true }, (_, filename) => {
+        if (filename && !/node_modules|[/\\]\./.test(filename)) {
+          BrowserWindow.getAllWindows().forEach((bw) => bw.webContents.reloadIgnoringCache());
+        }
+      });
```
eliykat: Is this related to the keytar replacement, or is it an unrelated fix? (either is fine, particularly because this never seems to work properly the first time the app compiles - just asking for context)

libs/services/state-service/stateMigration.service.ts
```diff
@@ -128,6 +136,27 @@ export class StateMigrationService {
...
+        `${clientId}_oneLoginClientSecret`,
+        `${clientId}_accessToken`,
+        `${clientId}_refreshToken`,
+        `${clientId}_twoFactorToken`,
+      ];
+      await Promise.all(
+        v3KeytarKeys.map((key) =>
+          passwords.migrateKeytarPassword(SECURE_STORAGE_SERVICE_NAME, key),
+        ),
+      );
```
eliykat: This appears to duplicate the `5To6` migration that you've added below. It's clearer to add a new migration rather than editing the old one.

libs/services/state-service/stateMigration.service.ts
```diff
@@ -207,6 +236,64 @@ export class StateMigrationService {
...
+        `${clientId}_gsuitePrivateKey`,
+        `${clientId}_azureKey`,
+        `${clientId}_entraIdKey`,
+        `${clientId}_entraKey`,
+        `${clientId}_oktaToken`,
+        `${clientId}_oneLoginClientSecret`,
+        `${clientId}_accessToken`,
+        `${clientId}_refreshToken`,
+        `${clientId}_twoFactorToken`,
+      );
```
eliykat: Why are we including these? We should rely on the new keys that we migrated to in the previous migration.

src-gui/services/electron/electronRendererSecureStorage.service.ts
```diff
@@ -1,43 +1,48 @@
...
+
 export class ElectronRendererSecureStorageService implements StorageService {
-  async get<T>(key: string): Promise<T> {
-    const val = ipcRenderer.sendSync("keytar", {
-      action: "getPassword",
-      key: key,
-      keySuffix: "",
-    });
-    return Promise.resolve(val != null ? (JSON.parse(val) as T) : null);
+  async get<T>(key: StorageKey | SecureStorageKey, options?: StorageOptions): Promise<T> {
```
eliykat: `storageOptions` has reappeared here 🔨

src-gui/services/electron/electronRendererSecureStorage.service.ts
eliykat, line 1: If the renderer can invoke the native code directly, can it just reuse the `nativeSecureStorageService`? The implementations look very similar.

package.json
eliykat, line 106: (non-blocking) You removed the call to electron-reload in the app code - do we still need it as a dependency?

libs/services/nativeSecureStorage.service.ts
```diff
@@ -0,0 +1,41 @@
...
+
+import { passwords } from "dc-native";
+
+export class NativeSecureStorageService implements StorageService {
+  constructor(private serviceName: string) {}
+
+  async get<T>(key: StorageKey | SecureStorageKey): Promise<T> {
+    return passwords
+      .getPassword(this.serviceName, key)
+      .then((val: string) => JSON.parse(val) as T)
```
eliykat: Are existing values stored as JSON strings? I can see new values get converted to JSON when saved, but I'm a bit wary of hitting an existing invalid JSON string if that's not already the current practice (or if there is some other corruption). I suggest some error handling - maybe returning an empty string and logging the issue to console (but making sure not to log any sensitive data).

native/src/migration.rs
eliykat, line 1: Does this have any equivalent in `clients`? I remember some mention of this being an issue, so good job picking it up. But I can't find any reference for this logic, and it's a bit over my head.

eliykat (changes requested):
Great job on this - excited to see these PRs getting merged. Comments below.

I also had Claude compare this implementation to the desktop app that it's based on. The feedback seemed good so here's the copy/paste. I think 5 is a red herring but left it in for completeness.

> Issues found
> 
>   1. Migration loop calls migrateStateFrom3To5() twice (minor correctness issue)
> 
>   migrate.ts:43-55: The while loop increments a local variable currentStateVersion by 1 each iteration. Starting from v3: runs migrateStateFrom3To5() at v3, increments to 4, runs it again at v4 (which is guarded and no-ops since the DB
>   now reads v5), then runs migrateStateFrom5To6() at v5. This is harmless but surprising — the guard at stateMigration.service.ts:66-68 saves you.
> 
>   2. migrateStateFrom5To6 migrates keys that are in regular storage, not the keychain
> 
>   stateMigration.service.ts:252-268: The v5→v6 migration calls migrateKeytarPassword on SecureStorageKeys.userDelta, groupDelta, lastUserSync, lastGroupSync. But default-state.service.ts:329-376 shows these are actually stored in
>   storageService (lowdb), not secureStorageService (keychain). They'd never exist in Credential Manager, so migrateKeytarPassword returns false immediately on all four — harmless no-ops, but misleading.

(I thought we'd fixed that one in a previous PR? oh well)

>   3. No macOS arm64 CLI binary in CI
> 
>   build.yml:124-182: The macos-cli job runs on macos-15-intel and only builds/zips dc_native.darwin-x64.node. The native/index.js has an arm64 case — if a user on Apple Silicon tries to run the CLI natively (not under Rosetta), it
>   throws "Could not find dc-native binary for darwin-arm64". This needs a separate arm64 build or a universal build, or explicit documentation that x64 (Rosetta) is the supported path.
> 
>   4. Dead loadError variable in native/index.js
> 
>   native/index.js:6,63-65: loadError is declared but never assigned. The check if (loadError) at line 63 is unreachable dead code. Copied from the desktop module but not fully adapted.
> 
>   5. pkg + __dirname for native module — needs verification
> 
>   webpack.cli.cjs:47 sets node: { __dirname: false }, which means webpack preserves __dirname at the module's location. Since dc-native is external (via nodeExternals()), pkg bundles node_modules/dc-native/index.js into the snapshot. In
>    pkg's virtual snapshot, __dirname for bundled code points to the virtual path, not the executable directory. The existsSync(join(__dirname, 'dc_native.linux-x64-gnu.node')) check could fail at runtime because the .node file isn't in
>   the virtual snapshot.
> 
>   The CI version test (bwdc -v) does implicitly test this — if the test passes in CI, it works. But the reason it works may depend on @yao-pkg/pkg v5's specific behavior around native addon detection (it may detect the require(filePath)
>    call and extract the .node file). This should be verified explicitly or documented.
> 
>   6. windows-gui build has no build:native step
> 
>   build.yml:311: The windows-gui job calls npm run dist:win which resolves to build:dist → rebuild → build:native:release. This is correct. But dist:win:ci (the CI pack variant) calls npm run build && npm run pack:win:ci, which does NOT
>    call rebuild. If CI ever uses dist:win:ci, the native module won't be built. Currently dist:win is used, so it's fine.
> 
>   ---
>   Minor observations
> 
>   - SecureStorageKeys.userDelta/groupDelta/lastUserSync/lastGroupSync are named as if they're in secure storage but are actually read/written via storageService in default-state.service.ts. The naming is confusing but internally
>   consistent.

BTreston:
@eliykat For your initial comment:
`#1` We probably don't have to do anything here. Its awkward but its documented and just an unfortunate quirk from the need to have a v3to5 migration bc of the error in the original implementation
`#3`I am already taking care of this in another PR for after this stack ([[DC] macOS Rosetta Support Ending](https://bitwarden.atlassian.net/browse/PM-33578)) I was also going to address `#6` here too, but I have no strong feeling about skipping it or doing it here.
I'll leave the implementation of the rest here to @JaredScar pls lmk if you need any help 👍

src-gui/main.ts
JaredScar: Unrelated. We used `electron-reload` before to watch for file changes in dev mode, but this replaces that. A better hot-reload watcher fix essentially

libs/services/state-service/stateMigration.service.ts
JaredScar: This accounts for old keys written via Keytar. The 5-6 migration covers the same old keys too under Keytar

package.json
JaredScar: Answered above

native/src/migration.rs
JaredScar: Yes, it was already solved with the clients repo looking back at Git history. It seems the migration piece of code was removed in a PR at a later date when enough time was deemed to have passed in order to remove the migration code.

https://github.com/bitwarden/clients/pull/2635/changes#diff-46b6e4f0b194b3add474aae52269f833fce555118c6b4530ea2d8d839f6d537e is where the values were taken from Keytar and updated the entry to the proper format going forward in the clients repo.

https://github.com/bitwarden/clients/pull/11900/changes#diff-ce4037ceaf5d4efc292be74b262cbb88022d8e913160cb07c340391f55b41b15L10 is when it got removed since I guess enough time had been deemed passed

libs/services/state-service/stateMigration.service.ts
BTreston: @eliykat  If I'm understanding this corretly, the reason we're coming back to this is because the next release will no longer have keytar and this migration has not been run in prod yet. I think it might actually be better to just do this all in this `v3To5` migration instead of migrating potentially incorrectly encoded data and then running another migration to fix that. We should probably just do the conversion here in one migration step and drop the `v5To6` migration altogether.

libs/models/state.model.ts
```diff
@@ -39,6 +39,12 @@ export const StorageKeys = {
   // TODO: Remove when jslib cleanup PR is merged
   appId: "appId",
   anonymousAppId: "anonymousAppId",
+
+  // Sync metadata (non-sensitive, stored in regular storage)
```
eliykat: (non-blocking) please catch these unnecessary Claude comments.

package.json
```diff
@@ -93,7 +95,7 @@
     "@types/tldjs": "2.3.4",
     "@typescript-eslint/eslint-plugin": "8.57.1",
     "@typescript-eslint/parser": "8.57.1",
-    "@yao-pkg/pkg": "5.16.1",
+    "@yao-pkg/pkg": "^6.14.2",
```
eliykat: We pin all our dependencies - this means no `~` or `^` which allow for version ranges.
And it should be pinned to exactly `yao-pkg/pkg@6.5.1` because this is the version we've vetted for node 22.

package.json
eliykat: (non-blocking) You did answer my question about this [above](https://github.com/bitwarden/directory-connector/pull/1012#discussion_r3052640844), but - can we remove it if we're no longer using it? Can be done in a follow up PR if more convenient.

EDIT: in fact, I would prefer to just revert to using `electron-reload` for now - a better reloader is welcome but outside the scope of these changes. Please open a separate PR.

libs/services/nativeSecureStorage.service.ts
```diff
@@ -1,16 +1,29 @@
...
 
 import { SecureStorageKey, StorageKey } from "../models/state.model";
 
 import { passwords } from "dc-native";
 
 export class NativeSecureStorageService implements StorageService {
-  constructor(private serviceName: string) {}
+  constructor(
+    private serviceName: string,
+    private logService?: LogService,
```
eliykat: We shouldn't have nullable dependencies like this. Why can't we inject logService in all cases?

libs/services/state-service/stateMigration.service.ts
eliykat: @JaredScar Yes exactly - both migrations handle the old keys under keytar, so they are duplicative. It should be in 1 migration or the other.

@BTreston I think the main advantage of separate migrations is that you have a separation of concerns:

- 3-5 migrates to the flat storage structure: it moves the data but does not change it
- 5-6 migrates away from keytar: it changes the data but does not move it

They'll run back-to-back either way (invisible to the user), but if one fails you have a better idea of what went wrong, and you're more likely to be in a recoverable state. If it's all in 1 and something fails, the possible causes and possible resulting states are much wider.

Also: why do we have 3-5 in one migration? Where's 4? (done in the StateService PR but I can't recall now)

libs/services/state-service/stateMigration.service.ts
```diff
@@ -261,30 +262,12 @@ export class StateMigrationService {
...
       SecureStorageKeys.apiKeyClientSecret,
       SecureStorageKeys.twoFactorToken,
-      SecureStorageKeys.userDelta,
-      SecureStorageKeys.groupDelta,
-      SecureStorageKeys.lastUserSync,
-      SecureStorageKeys.lastGroupSync,
+      StorageKeys.userDelta,
+      StorageKeys.groupDelta,
+      StorageKeys.lastUserSync,
+      StorageKeys.lastGroupSync,
```
eliykat: @BTreston again, maybe something we resolved in the StateService PR, so sorry if this is a repeat: are these saved in secure storage today? If so, are we moving them to plaintext storage, and we need to handle that? Or is it just a mistake that they were ever included in the SecureStorageKeys? (if so they can just be removed from here)

native/src/migration.rs
eliykat: Thank you for those links! This code is my main concern with this PR, none of us are rust devs so I assume it's all Claude generated. It has some similarities to the code you linked ([here](https://github.com/bitwarden/clients/blob/2ce8500391fed075e2b55cbff0eded5371f114f7/apps/desktop/desktop_native/core/src/password/windows.rs) is the easiest to read, immediately before removal), but I am not confident as it is.

.github/workflows/build.yml
```diff
@@ -182,6 +182,67 @@ jobs:
           if-no-files-found: error
 
 
+  macos-arm64-cli:
```
eliykat: @JaredScar Do we have an arm64 build today? If not, I am OK with leaving it to @BTreston 's follow-up PR.

package.json
```diff
@@ -47,16 +47,18 @@
     "pack:win": "npm run clean:dist && electron-builder --win --x64 --ia32 -p never -c.win.certificateSubjectName=\"8bit Solutions LLC\"",
     "pack:win:ci": "npm run clean:dist && electron-builder --win --x64 --ia32 -p never",
     "pack:cli": "npm run pack:cli:win | npm run pack:cli:mac | npm run pack:cli:lin",
-    "pack:cli:win": "pkg ./src-cli --targets win-x64 --output ./dist-cli/windows/bwdc.exe",
-    "pack:cli:mac": "pkg ./src-cli --targets macos-x64 --output ./dist-cli/macos/bwdc",
-    "pack:cli:lin": "pkg ./src-cli --targets linux-x64 --output ./dist-cli/linux/bwdc",
+    "pack:cli:win": "pkg ./src-cli --targets node22-win-x64 --output ./dist-cli/windows/bwdc.exe",
+    "pack:cli:mac": "pkg ./src-cli --targets node22-macos-x64 --output ./dist-cli/macos/bwdc",
+    "pack:cli:mac:arm64": "pkg ./src-cli --targets node22-macos-arm64 --output ./dist-cli/macos-arm64/bwdc",
```
eliykat: Same here depending on the decision re: arm64. (and again below)

eliykat (changes requested):

native/src/migration.rs
```diff
@@ -0,0 +1,67 @@
...
+        .map(|s| !s.contains('\0'))
+        .unwrap_or(false);
+
+    if !blob_is_utf8 {
+        // Already UTF-16 or unrecognised format; no migration needed.
+        return Ok(false);
+    }
+
+    let utf8_value = String::from_utf8(blob_bytes).map_err(|e| anyhow!(e))?;
+    desktop_core::password::set_password(service, account, &utf8_value).await?;
```
Hinton: This is quite different from the migration logic in desktop? In desktop we left the fallback logic to the JS side in a catch. I.e. if the data was invalid, we tried decoding it as utf16.

That's not to say this code would not work, it probably would but it seems to duplicates what desktop_native does.

native/src/migration.rs
eliykat: OK, I did a bit more digging here, here's how we can reuse the same approach directly without writing any new rust code.

- copy the `get_password_keytar` implementation as is ([code ref](https://github.com/bitwarden/clients/blob/2ce8500391fed075e2b55cbff0eded5371f114f7/apps/desktop/desktop_native/core/src/password/windows.rs))
- loop through the storage keys, reading them with `get_password_keytar` and then setting them with `set_password` in order to migrate them. See https://github.com/bitwarden/clients/blob/78986023e779540a9c4ffe7952592c769a6560bd/src/main/desktopCredentialStorageListener.ts#L45-L55. We can probably use the same JSON parsing check as this does to detect old keys, but test it. If in doubt, we know our migration will only run once, so it's reasonable to assume that any key we encounter is in the old format.

libs/services/state-service/stateMigration.service.ts
BTreston: Makes sense to me, I'm fine with it.

In the original implementation, the migrate function was overridden in the non-jslib class, but the **implementation** to migrate to state 4 was only in the overridden jslib cass's def. During development and testing I was manually triggering the migrations, but it didn't quite sit right with me to do that. So I wrote a script to clean all local state (secure & non) from disk to test the full migration path. When I did this I discovered that the v4 migration was not actually running (b/c is was in the overridden function def). A shorter version of this story is documented in code, however I guess to clear this up maybe we can just rename: v5 => v4 and v6 => v5

package.json
BTreston: +1 to reverting. This is being done in the followup PR that cleans up our deps.

libs/services/state-service/stateMigration.service.ts
BTreston: They are not in secure storage today, this was wrong in the state service PR. This is fine IMO

.github/workflows/build.yml
BTreston: please do revert it, I am working on the node 25 SEA builds in a followup
[[DC] macOS Rosetta Support Ending](https://bitwarden.atlassian.net/browse/PM-33578)

src-gui/main.ts
```diff
@@ -49,11 +49,7 @@ export class Main {
     const watch = args.some((val) => val === "--watch");
 
     if (watch) {
-      fs.watch(__dirname, { recursive: true }, (_, filename) => {
-        if (filename && !/node_modules|[/\\]\./.test(filename)) {
-          BrowserWindow.getAllWindows().forEach((bw) => bw.webContents.reloadIgnoringCache());
-        }
-      });
+      electronReload(__dirname, {});
```
eliykat: Not properly reverted - check the diff. The inline import is required so that it is only loaded in dev mode.

libs/services/state-service/stateMigration.service.ts
eliykat: These keys are not in secure storage and so do not need to be migrated here. These lines can be deleted.

native/src/migration.rs
```diff
@@ -0,0 +1,57 @@
...
+            (*credential).CredentialBlobSize as usize,
+        ))
+    };
+
+    Ok(String::from(password))
+}
+
+pub async fn migrate_keytar_password(service: &str, account: &str) -> Result<bool> {
+    let value = match get_password_keytar(service, account) {
+        Err(_) => return Ok(false),
```
eliykat, line 50: It would be better to return the error, but let's do that in a smaller follow-up PR so as not to block this one.

eliykat (changes requested):
Great work! Just 2 things to action below (do not do the rust change in this PR).

Also:
- lint is failing
- some builds are failing, they look like they still depend on keytar. e.g. https://github.com/bitwarden/directory-connector/actions/runs/24210945042/job/70679266955?pr=1012

JaredScar:
As a note, the builds will fail via the GitHub checks because it pulls in the `build.yml` from the `main` branch rather than using the existing yml on this branch with the valid changes

eliykat (approved):

michalchecinski (approved):

PR #1048, opened 2026-03-16
[PM-33767] fix: validate organizationId in bwdc test to surface missing config early

### LINKED ISSUE #141: bwdc test does not check for the existence of an organizationId
If `organizationId` is missing from an otherwise complete `data.json`, then:

- `bwdc test` will work, correctly showing users and groups that should be synced.
- `bwdc sync` will fail, returning `Organization not set.`

`bwdc test` should identify this case, otherwise it's not very useful for testing configuration changes.
This was tested on v2.9.3 against an Azure Active Directory and a Bitwarden Cloud Enterprise org.

sputnik-mac:
## Problem

Fixes #141

When `organizationId` is missing from `data.json`, `bwdc test` completes successfully showing users and groups, but `bwdc sync` fails with `Organization not set.`. This makes `bwdc test` misleading — a passing test implies the configuration is ready for sync, but it is not.

## Root Cause

The `organizationId` validation only occurs in `generateHash()`, which is only called during a real sync (not in test mode). In test mode, `sync()` returns early before reaching this validation.

## Fix

Move the `organizationId` check to the beginning of `sync()`, before any directory queries are made. This ensures both `bwdc test` and `bwdc sync` fail fast with a clear `Organization not set.` error when the organization is not configured.

## Tests

Added two new unit tests:
- `Sync throws 'Organization not set.' when organizationId is missing`
- `bwdc test (test=true) throws 'Organization not set.' when organizationId is missing`

All existing tests continue to pass.

### DIFF
--- a/src/services/sync.service.spec.ts
+++ b/src/services/sync.service.spec.ts
@@ -68,6 +68,20 @@ describe("SyncService", () => {
     );
   });
 
+  it("Sync throws 'Organization not set.' when organizationId is missing", async () => {
+    stateService.getOrganizationId.mockResolvedValue(null);
+    stateService.getSync.mockResolvedValue(getSyncConfiguration({ groups: true, users: true }));
+
+    await expect(syncService.sync(true, false)).rejects.toThrow("Organization not set.");
+  });
+
+  it("bwdc test (test=true) throws 'Organization not set.' when organizationId is missing", async () => {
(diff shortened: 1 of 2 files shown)


JaredScar (changes requested):
Please resolve the conflicts then request re-review. Thanks!

PR #1050, opened 2026-03-18
[PM-11854] Add latest tools from template repo

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-11854

## 📔 Objective

Aims to modernize the DC repo to add useful tools from the template repo here.


r-tome (dismissed):

r-tome:
Claude picked up on these:

- **`prepare` script needs Husky 9 syntax**

DC's `package.json` has `"prepare": "husky install"` but Husky 9 just uses `"prepare": "husky"`. Since DC already has Husky 9.1.7 installed, this should be updated.

- **3Prettier `--cache` flag in lint-staged**

The template uses `--cache` for faster Prettier runs in `lint-staged`. Consider updating:

```json
"./!(jslib)**": "prettier --cache --ignore-unknown --write",

**Comment by BTreston** [comment]
@JaredScar pls address the linter errors. TY

**Review by BTreston — DISMISSED** [review]

**Review by theMickster — DISMISSED** [review]
Approving the Claude changes. The rest are left to the team

File: .editorconfig [hunk 1]
```diff
@@ -3,13 +3,27 @@
 # top-most EditorConfig file
 root = true
 
-# Unix-style newlines with a newline ending every file
+# Don't use tabs for indentation.
```
withinfocus, line 6: ⚠️ This caused a lot of changes here -- CJS, CSS, MD, etc. and not sure it's tuned to the repo.

.github/renovate.json5
```diff
@@ -1,24 +1,34 @@
...
+        {
+            groupName: "github-action minor",
+            matchManagers: ["github-actions"],
+            matchUpdateTypes: ["minor", "patch"],
+        },
+        {
+            groupName: "npm minor",
+            matchManagers: ["npm"],
+            matchUpdateTypes: ["minor", "patch"],
+        },
```
eliykat: We don't group npm minor dependencies - in the past it's made for big unwieldy renovate PRs. We've had more success with more, smaller PRs. We can review this decision if needed but I assume this change is not really intentional.

Recommendation: remove this npm minor grouping. And the docker-compose minor grouping as well. (Keep it in `enabledManagers` though, it is good to start tracking it.)

.cspell.json
eliykat, line 1: cspell is a spellchecker, but we don't use it in this repo. This can be removed.

.editorconfig
eliykat: Agree, I think we need to align this a bit more with current practice. MD changes make sense to me, it's good to have hard line wrapping. But cjs and mjs files are just JS and should be included in the JS block for consistency. Similarly, CSS should follow the same rules as SCSS. That will probably clean up most of the noise.

eliykat (changes requested):

eliykat (commented):
My comments have been addressed but the css/mjs/cjs files need their formatting updated/reverted to match (i.e. they should drop off the PR diff, because the new settings should match current practice).

BTreston (dismissed):

eliykat (approved):

PR #1051, opened 2026-03-20
[BEEEP] - Sign and notarize CLI build binary

BTreston:
## 📔 Objective
Adds access to secrets to sign the cli build binary for the macos-cli build step to resolve the error below:
image

The build artifact in https://github.com/bitwarden/directory-connector/actions/runs/23356740801/job/67949409099?pr=1052 has the now working binary.

PR #1077, opened 2026-04-07
[PM-33578] Mac OS arm builds

BTreston:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-33578

## Summary

This PR adds native macOS ARM64 (Apple Silicon) support for both the CLI and GUI builds. It also replaces `pkg` with Node.js SEA (Single Executable Applications).

### Key Changes

**macOS ARM64 Support**
- Add new `macos-arm64-cli` and `macos-arm64-gui` CI jobs running on `macos-15` (Apple Silicon runners)
- Add `pack:cli:mac:arm64`, `dist:cli:mac:arm64`, `dist:mac:arm64`, and `pack:mac:arm64` npm scripts
- Update `electron-builder.json` to use `${arch}` in artifact names so x64 and ARM64 artifacts are distinguishable (e.g. `Bitwarden-Connector-X.Y.Z-arm64.dmg`)
- Add JIT entitlement to `resources/entitlements.mac.plist` required for Apple Silicon code signing

**CLI Packaging: `pkg` → Node.js SEA**
- Replace `pkg` with a custom `scripts/pack-sea.mjs` script using Node.js SEA (Single Executable Applications)
- Downloads official Node.js binaries (version-matched, cached in `.node-sea-base/`) as SEA base to avoid issues with Homebrew/nvm-installed binaries
- **macOS x64 exception**: Due to a known upstream Node.js issue ([nodejs/node#59553](https://github.com/nodejs/node/issues/59553)) where Mach-O injection corrupts TLS metadata or causes codesign failures on Intel Macs, macOS x64 ships a shell-wrapper bundle instead of a single binary: `bwdc` (launcher script), `node`, `bwdc.js`, and `*.node` native addon
- macOS ARM64, Linux, and Windows use true single-executable binaries
- Update CI zip steps to use directory-based zipping (`dist-cli/<platform>/*`) to include all bundle files


.github/workflows/build.yml
```diff
@@ -53,6 +53,7 @@ jobs:
     runs-on: ubuntu-24.04
     needs: setup
     env:
+      NODE_OPTIONS: --max_old_space_size=4096
```
eliykat: If this is needed here, it's probably better to specify it in the root `.npmrc` so that all scripts/workflows get it: e.g. https://github.com/bitwarden/clients/blob/d2905e30a9c772e3bef83f7db7c8a7feb6ae9d0c/.npmrc#L2-L4

If 1 build script is running out of memory, it's probably only a matter of time until others do as well, and it saves you having to specify it everywhere.

scripts/pack-sea.mjs
```diff
@@ -0,0 +1,164 @@
...
+  const archivePath = join(officialNodeCache, archiveName);
+
+  console.log(`Downloading official Node.js ${nodeVersion} binary...`);
+  console.log(`  URL: ${downloadUrl}`);
+
+  execFileSync("curl", ["-fsSL", "-o", archivePath, downloadUrl], {
+    stdio: "inherit",
+  });
+
+  console.log("Extracting node binary...");
```
eliykat: Please add checksum verification to protect against supply chain attacks
(EDIT: if this is still needed - see next comment)

scripts/pack-sea.mjs
```diff
@@ -0,0 +1,164 @@
+#!/usr/bin/env node
+/**
+ * Pack the CLI binary using Node.js SEA (Single Executable Applications).
+ * Requires Node 25.5+ for the --build-sea flag.
+ *
+ * Uses the official Node.js binary (downloaded if needed) as the base, since
+ * package-manager-installed Node binaries (e.g. Homebrew) may lack the SEA
+ * fuse sentinel required by --build-sea.
```
eliykat: This approach broadly looks fine, but is it necessary?

This comment refers to the "SEA fuse sentinel", but the [docs](https://nodejs.org/api/single-executable-applications.html#-build-sea) indicate that the fuse sentinel is something that's set in the binary artifact, not in the system's node runtime, which is what you're concerned about here.

Most of this script is just about downloading a specific official version of node. But my assumption is that we can already do that - via `nvm` + `.nvmrc` for local development, or via the `setup-node` action in CI workflows.

scripts/signal-exit-shim.cjs
```diff
@@ -0,0 +1,6 @@
+// Shim: proper-lockfile expects require('signal-exit') to return the onExit function
+// directly (signal-exit v3 behavior). signal-exit v4 uses named exports only.
+// This shim re-exports the function as the module default.
```
eliykat: Can you help me understand this? Presumably `proper-lockfile` depends on `signal-exit` v3, not v4; why are we using v4 with a shim instead of v3? 

I asked Claude and it said that webpack flattens dependencies, but that would be pretty drastic behavior so I'm not sure about that answer - but happy to be wrong if this is what you've verified. If that is the case - is there any way to override this and package both?

It would be better to use the correct dependency version, as we don't know what other breaking changes there may be.

package.json
```diff
@@ -45,19 +45,21 @@
     "clean:dist:cli": "rimraf --glob ./dist-cli/*",
     "pack:lin": "npm run clean:dist && electron-builder --linux --x64 -p never",
     "pack:mac": "npm run clean:dist && electron-builder --mac -p never",
-    "pack:win": "npm run clean:dist && electron-builder --win --x64 --ia32 -p never -c.win.certificateSubjectName=\"8bit Solutions LLC\"",
```
eliykat, line 48: Do we need to pass the `c.win.certificateSubjectName` here as well? (not familiar with what exactly this does, just checking)

eliykat (changes requested):
This is awesome, great work.

scripts/pack-sea.mjs
BTreston: yep, good point. Forgot about that node step in CI.

scripts/signal-exit-shim.cjs
BTreston: afaik its purely a transitive dependency b/c `proper-lockfile` ships their own v3 bundled into their app, so we can't control it. Other deps bundle v4 so we can't blanket change over to v3. But I think the better alternative is aliasing directly to the version in `proper-lockfile`s `node_modules` maybe. I'll see if I can get it working without the this "shim" as I was already not a fan of this.

package.json
BTreston: We don't need it if we're using a custom sign script (which we are, `sign.mjs`) This is also true for `publish:win` below, so I've removed it from there too

webpack.cli.mjs
```diff
@@ -66,16 +65,29 @@ const config = {
     plugins: [new TsconfigPathsPlugin({ configFile: "./tsconfig.json" })],
     symlinks: false,
     modules: [path.resolve("node_modules")],
+    alias: {
+      // proper-lockfile requires signal-exit v3 (default export = onExit function).
+      // The top-level signal-exit is v4 (named exports only), but proper-lockfile
+      // ships its own nested v3. Point webpack directly at that copy.
+      "signal-exit": path.resolve(
+        __dirname,
+        "node_modules/proper-lockfile/node_modules/signal-exit/index.js",
```
BTreston: @eliykat I think this ends up being a global alias (including for deps that expect v4) but the build pass in CI and the app seems fine. I didn't think this would work but apparantly it does, but I'm not sure why. I'll follow up if i figure it out.

edit: this breaks login functionality (b/c thats where we use inquirer, which uses v4 I believe)

scripts/pack-sea.mjs
```diff
@@ -0,0 +1,84 @@
...
+
+  if (platform === "macos-arm64") {
+    console.log("Ad-hoc signing binary...");
+    // --force is required because the official Node.js binary already carries an Apple
+    // signature; without it codesign refuses to replace an existing signature (see Apple
+    // TN2206 "Using the codesign Tool" – https://developer.apple.com/library/archive/technotes/tn2206/_index.html).
+    execFileSync("codesign", ["--sign", "-", "--force", outputBinary], {
+      stdio: "inherit",
+    });
+  }
```
brandonbiete, line 89: Do the native modules need to be code signed? The `bwdc` binary gets signed (line 76), but the native `.node` modules that get copied to the output directory (lines 64-69) don't appear to be signed.
On macOS, unsigned binaries can potentially trigger Gatekeeper warnings, something like "cannot be opened because the developer cannot be verified" etc.

BTreston: Some of this was going to be handled in https://github.com/bitwarden/directory-connector/pull/1051, but I've decided to roll it into this PR because they fit nicely together. 

The native module did indeed need to be signed, the latest changes should have that set now. The build artifacts in https://github.com/bitwarden/directory-connector/actions/runs/24476942691/job/71531511465?pr=1078 show this working

eliykat (dismissed):

eliykat (approved):

PR #1088, opened 2026-04-13
[PM-35035] [Fix] Combine includeGroup with excludeGroup and exclude clauses in E…

Joly0:
…ntra ID user filter, document syntax in UI

Previously, an Entra ID user filter like "includeGroup:A|excludeGroup:B|exclude:user@x.com" silently dropped every clause after the first, so include and exclude filters could never be combined as an intersection. Therefore this PR fixes this so when includeGroup is the primary clause, it parses all secondary excludeGroup and exclude clauses (separated by | ) and subtract them from the included set.

Also adds a help icon next to the User Filter label in Settings, showing per-directory filter syntax documentation as a hover tooltip, since this syntax was previously undocumented in the UI.

## 🎟️ Tracking
https://github.com/bitwarden/directory-connector/issues/1067

## 📔 Objective

After being annoying by this bug and the issue being closed and the support that couldnt help me with the problem cause it was not a configuration issue but a bug in the software i looked myself through the code and figured that this was indeed a bug. 

## 📸 Screenshots

Settings before:
grafik

Settings after:
grafik

Test sync before (exclude filters are completly ignored):
grafik

Test sync after (exclude filtes are properly applied on top of the includeGroup filter):
grafik


Joly0:
Hey, just wanted to ask if this will get a review anytime soon?

libs/services/directory-services/entra-id-directory.service.ts
r-tome, line 269: The [new help text](https://github.com/bitwarden/directory-connector/pull/1088/changes#diff-9596edb02fb61dd2b15cd0b4c8ffad60da970d8236ed80f4e4cffed41e15916fR418) says clauses can be separated by `|` or a newline, but this parser only splits on `|`

r-tome (commented):
Thanks for contributing and being so patient!
This doesn't seem like a bug to me, [the documentation](https://bitwarden.com/help/microsoft-entra-id/) doesn't demonstrate combining clauses so its more of a feature request. It's a reasonable one so I have brought this up internally to see if we want to support it. I'll circle back once I hear back.

libs/services/directory-services/entra-id-directory.service.ts
Joly0: Ah, my bad. While testing/implementing this I had it so newlines would also worked, but in the end decided against it. But apparently left the help text in there, which of course is wrong. I can remove that (or you if this would get merged).

Joly0:
> Thanks for contributing and being so patient!

Thanks for taking the time and looking into it. I highly appreciate it :)

> This doesn't seem like a bug to me, [the documentation](https://bitwarden.com/help/microsoft-entra-id/) doesn't demonstrate combining clauses so its more of a feature request. It's a reasonable one so I have brought this up internally to see if we want to support it. I'll circle back once I hear back.
> 
Yes, you are correct, the way I implemented it is basically a new feature/a feature request, but it also solves a problem I would consider a bug. The problem is, that various filters overwrite each other, which the documentation doesn't mention which I therefore consider as a bug (in my humble opinion, but I totally see your point).

You could argue that this is just lack of information in the documentation, but as I read it `You can include or exclude users from a sync based on their Microsoft Entra ID group membership using the includeGroup **AND** excludeGroup` I understood it as if this was or should be possible, but it wasn't.

libs/services/directory-services/entra-id-directory.service.ts
Joly0: Hey, so i looked through the code again and while the secondary parser did parse on newlines aswell, the primary didnt. So i changed that part and the help test to actually reflect the helptext, so both are correct now.

From my side this should be resolved now.

r-tome:
> You could argue that this is just lack of information in the documentation, but as I read it You can include or exclude users from a sync based on their Microsoft Entra ID group membership using the includeGroup **AND** excludeGroup I understood it as if this was or should be possible, but it wasn't.

Fair point, the documentation is ambiguous in that sense and could be clearer. I've provided that feedback internally.

I raised internally whether we want to support combining clauses at all and haven't heard back yet. I've bumped it again and will let you know as soon as I do.

Joly0:
Appreciate it. I would really like seeing this being supported in the future :D

Joly0:
Hey, was wondering if there is any update regarding my pull request? :)

libs/services/directory-services/entra-id-directory.service.ts
```diff
@@ -94,6 +94,18 @@ export class EntraIdDirectoryService extends BaseDirectoryService implements IDi
     // Only get users for the groups provided in includeGroup filter
     if (setFilter != null && setFilter[0] === UserSetType.IncludeGroup) {
       users = await this.getUsersByGroups(setFilter);
+      // If a secondary excludeGroup clause is also present in the filter
+      // string (e.g. "includeGroup:A|excludeGroup:B"), subtract the members
+      // of those groups from the included set so include and exclude can be
+      // combined as an intersection rather than mutually exclusive options.
+      const secondaryExcludeGroup = this.extractSecondaryExcludeGroupSet(
```
r-tome, line 101: Could you please add tests for this new include/exclude combining logic? Since this decides which users get synced, I'd like to guard against regressions. We already have some EntraID integration tests in https://github.com/bitwarden/directory-connector/blob/main/libs/services/directory-services/gsuite-directory.service.integration.spec.ts. It could be worth covering different configurations: for example `includeGroup:A|excludeGroup:B` and `includeGroup:A|exclude:user@x.com`

libs/services/directory-services/entra-id-directory.service.ts
```diff
@@ -275,6 +303,103 @@ export class EntraIdDirectoryService extends BaseDirectoryService implements IDi
...
+    }
+
+    if (set.size === 0) {
+      return null;
+    }
+    return [UserSetType.ExcludeGroup, set];
+  }
+
+  // Parses any secondary "exclude:<email>" clauses out of a user filter
+  // string. Mirrors extractSecondaryExcludeGroupSet but collects user
```
r-tome: As you've stated here these methods have a lot of duplicated logic. Could we extract that into a shared helper and have both methods call it?

r-tome (changes requested):
Hey! I have some news... this got the green light internally! 🎉 

The logic looks good to me. I just have a couple requests below.

Also, heads up that this is behind `main` so it needs a rebase. I noticed that because the EntraID integration tests aren't in this branch.

Joly0:
Awesome, glad its been green-lighted.
I will look into the requested changes as soon as possible, should'nt take too long. Will update in the next couple of days (probably).
Thank you very much :)

libs/services/directory-services/entra-id-directory.service.ts
Joly0: Ok, done. `extractSecondaryClauseValues` is now the shared helper, contains all the logic that used to be duplicated. `extractSecondaryExcludeGroupSet` and `extractSecondaryExcludeUserSet` are just wrappers around the first function now.

libs/services/directory-services/entra-id-directory.service.ts
Joly0: I have added unit tests with a mocked Graph client covering exactly the configurations you listed: includeGroup:A|excludeGroup:B (single and multiple excludeGroups), includeGroup:A|exclude:user@x.com (single, comma-separated, case-insensitive), and both combined in one filter. I saw the Entra integration spec that came in with #1140. I could extend it with a combining case against the seeded tenant (e.g. includeGroup:GroupA|exclude:amarshall@bwrox.dev), but since I cant access the shared .env credentials to verify locally, let me know if you'd like me to add that for CI to validate.

r-tome: Yes, please add the integration test. I'll approve the workflow to run the tests on CI

src-gui/app/tabs/settings.component.html
```diff
@@ -661,7 +661,17 @@ <h3 class="card-header">{{ "sync" | i18n }}</h3>
         </div>
         <div [hidden]="!sync().users">
           <div class="mb-3">
-            <label for="userFilter" class="form-label">{{ "userFilter" | i18n }}</label>
+            <label for="userFilter" class="form-label">
+              {{ "userFilter" | i18n }}
+              @if (directory != null) {
```
r-tome: This compares the signal function so it's never null.
```suggestion
              @if (directory() != null) {
```

r-tome (changes requested):

Joly0:
I implemented the requested integration test. Please let me know if any more changes are required or if the tests don´t work as expected.

src-gui/app/tabs/settings.component.ts
```diff
@@ -92,6 +92,23 @@ export class SettingsComponent implements OnInit, OnDestroy {
...
+        return this.i18nService.t("userFilterHelpOkta");
+      case DirectoryType.GSuite:
+        return this.i18nService.t("userFilterHelpGSuite");
+      case DirectoryType.OneLogin:
+        return this.i18nService.t("userFilterHelpOneLogin");
+      default:
+        return "";
+    }
+  }
+
```
BTreston: Can we convert this into a computed signal? Calling functions in template binding is an anti-pattern in angular

BTreston (changes requested):
This looks good, just one small nit ⛏️ 

Thank you for adding the tests, much appreciated!

src-gui/app/tabs/settings.component.ts
Joly0: Done. Please let me know if any other changes need to be done :)

BTreston (approved):
Looks good to me, thank you! I'll let @r-tome handle it from here 👍

r-tome (approved):

r-tome:
@Joly0 this has been merged. It will now go through QA testing and will be part of our next Directory Connector release.

Thank you so much for contributing, for being patient through this whole process, and for being open to our requests. Really appreciate it! 😀

Joly0:
I have to thank you @r-tome for being so open to my PR and appreciate the time that you and @BTreston invested to get this PR ready and merged.

Hopefully everything will go smoothly through QA :D

If any questions or problems arise, please let me know, I really want this feature to be useful :)

Really looking forward for the final release with this feature included

PR #1105, opened 2026-04-30
[PM-36089] fix windows gui build

BTreston:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-36089?issueKey=PM-36089&subProduct=jira-software

## 📔 Objective
fix file path resolution to use `__dirname` (the ESM drop in replacement, not the CJS version) to fix platform specific file path issues.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,6 +6,7 @@ on:
       - "main"
       - "rc"
       - "hotfix-rc"
+      - "test-windows-build"
   workflow_dispatch: {}
   workflow_call: {}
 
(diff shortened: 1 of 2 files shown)


eliykat (commented):
If it works it works, but - isn't `__dirname` CJS and `import.meta` is ESM? Does this mean this module is being run as CJS? It may not matter for this one, but it's odd.

BTreston:
> If it works it works, but - isn't `__dirname` CJS and `import.meta` is ESM? Does this mean this module is being run as CJS? It may not matter for this one, but it's odd.

@eliykat we define `__dirname` in this file to use `import.meta.url` as a drop in replacement for the CJS `__dirname`. We can change it to use `import.meta.url` inline, but imo I don't think it matters all that much.

eliykat (approved):
Oh good point - I didn't realise it was a defined local variable. Thanks.

PR #1034, opened 2026-03-12
[PM-31850] Enable ESM, migrate remaining files

BTreston:
🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-31850

## 📔 Objective

Converts webpack configs from `.cjs` to `.mjs`, enables ESM by adding `"type": "module"` to `package.json`, upgrades ESM-incompatible dependencies to their ESM-only versions, and add tests for new state version.

## Removed Dependencies

**Production**
- `big-integer` — removed
- `node-fetch` — removed (native fetch used instead)
- `parse5` — removed
- `tldjs` — removed
- `uuid` — removed

**Dev**
- `electron-reload` — removed
- `@types/lowdb` — removed (lowdb now ships its own types)
- `@types/node-fetch` — removed (along with node-fetch)
- `@types/tldjs` — removed (along with tldjs)
- `form-data` (dev duplicate) — removed
- `glob` — removed

## Major Version Bumps

| Package | Before | After |
|---|---|---|
| `chalk` | 4.1.2 | 5.6.2 (ESM-only) |
| `inquirer` | 8.2.6 | 13.3.2 (ESM-only) |
| `lowdb` | 1.0.0 | 7.0.1 (ESM-only) |
| `https-proxy-agent` | 7.0.6 | 8.0.0 |
| `electron` | 39.2.1 | 41.0.3 |
| `electron-builder` | 24.13.3 | 26.8.1 |
| `electron-store` | 8.2.0 | 11.0.2 |
| `@yao-pkg/pkg` | 5.16.1 | 6.5.1 |
| `copy-webpack-plugin` | 13.0.0 | 14.0.0 |
| `cross-env` | 7.0.3 | 10.1.0 |
| `webpack-cli` | 6.0.1 | 7.0.2 |
| `@electron/notarize` | 2.5.0 | 3.1.1 |
| `googleapis` | 149.0.0 | 171.4.0 |

## Minor/Patch Bumps

| Package | Before | After |
|---|---|---|
| `@angular/*` | 21.1.x | 21.2.x |
| `@angular-eslint/*` | 21.1.x | 21.3.x |
| `@angular/build` | 21.1.2 | 21.2.2 |
| `@angular/compiler-cli` | 21.1.1 | 21.2.4 |
| `@ngtools/webpack` | 21.1.2 | 21.2.2 |
| `@typescript-eslint/*` | 8.54.0 | 8.57.1 |
| `@types/inquirer` | 8.2.10 | 9.0.9 |
[...]

PR #1106, opened 2026-04-30
[PM-22352] Removes LowDB and replace with internal typings

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-22352

## 📔 Objective

Removes LowDB usage in codebase and utilizes internal typings instead

### DIFF
--- a/libs/services/baseLowdbStorage.service.ts
+++ b/libs/services/baseLowdbStorage.service.ts
@@ -1,11 +1,9 @@
 import * as fs from "fs";
 import * as path from "path";
 
-import { LowSync } from "lowdb";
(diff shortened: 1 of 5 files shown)


JimmyVo16 (approved):

eliykat (changes requested):
@JaredScar Please re-read the ticket - the scope here is not to replace `lowdb`, it's to replace its `@types/lowdb` package. `@types/` packages provide TS types for JS packages, but `lowdb` now provides its own TS types.

In this case it was already removed by https://github.com/bitwarden/directory-connector/pull/1034, so this PR and ticket can be closed.

PR #1162, opened 2026-07-09
test

BTreston:


BTreston:
Tested on ubuntu v22.04, no more error.

PR #1164, opened 2026-07-15
[PM-34488] directory connector not supporting multiple ad accounts on windows for multi directory sync

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34488

## 📔 Objective

Introduces a ID to configs for the directory secrets, so Secure Storage will get the secret along with the id rather than using a single shared key every time directory connector runs.

### DIFF
--- a/libs/models/ldapConfiguration.ts
+++ b/libs/models/ldapConfiguration.ts
@@ -1,6 +1,13 @@
 import { IConfiguration } from "./IConfiguration";
 
 export class LdapConfiguration implements IConfiguration {
+  /**
+   * Unique identifier for this specific configuration (generated automatically). Used to scope
+   * the secure-storage entry for `password` so that distinct configurations - e.g. two AD service
+   * accounts used for multi-directory sync - never share (or overwrite) each other's secret, even
+   * if their `data.json` files are swapped in and out on the same machine.
+   */
+  id?: string;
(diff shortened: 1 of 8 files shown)


JimmyVo16 (approved):

libs/services/state-service/secret-storage-key.util.ts
BTreston, line 1: @JaredScar Does this need a state migration? Existing secrets wont be prefixed by anything, how does this fix the issue for existing data?

libs/models/ldapConfiguration.ts
```diff
@@ -1,6 +1,13 @@
 import { IConfiguration } from "./IConfiguration";
 
 export class LdapConfiguration implements IConfiguration {
+  /**
+   * Unique identifier for this specific configuration (generated automatically). Used to scope
+   * the secure-storage entry for `password` so that distinct configurations - e.g. two AD service
+   * accounts used for multi-directory sync - never share (or overwrite) each other's secret, even
+   * if their `data.json` files are swapped in and out on the same machine.
+   */
+  id?: string;
```
BTreston, line 10: If the goal is to link a configuration to a specific secret by this guid then it should not be optional. If we need to make any further changes to this pls loop me in on the next PR

PR #1194, opened 2026-07-22
Allow customer-only Google Workspace config

### LINKED ISSUE #1193: [DC] Google Workspace: users and groups on secondary domains silently excluded when both Domain and Customer ID are set; customer-only config rejected
### Steps To Reproduce

1. Have a Google Workspace account with a secondary domain (primary `example.com`, secondary `example-two.com`), with at least one user and one group whose addresses are on the secondary domain.
2. In Directory Connector, configure the Google Workspace directory service and fill in BOTH the `Domain` field (`example.com`) AND the `Customer ID` field (the account's customer ID), the settings screen offers both fields side by side.
3. Run a test sync ("Test Now" in the desktop app, or `bwdc test` in the CLI).
4. Observe the returned users and groups.
5. Additionally: clear the `Domain` field, keep `Customer ID` set, and run the test sync again.

### Expected Result

- With `Customer ID` configured, the sync should return ALL users and groups across every domain of the Workspace account. Google's Directory API documentation recommends the `customer` parameter for multi-domain accounts: "if a customer has secondary domains, using the domain parameter only returns users with email addresses on that particular domain... we recommend using the customer parameter." [Source](https://developers.google.com/workspace/admin/reseller/v1/support/directory_api_common_errors#400badrequest)
- A customer-only configuration should be accepted as valid, since the Directory API requires exactly one of `domain` / `customer` — either should suffice.

### Actual Result

- When both fields are set, both parameters are sent to the Directory API, and the API gives `domain` precedence. Users AND groups on secondary domains are silently excluded from the sync, no warning or log message. Verified against a live multi-domain tenant: a test sync returned 7 of 8 users and 1 of 2 groups; the missing user and the missing group were exactly the ones addressed on the secondary domain.
- When only `Customer ID` is set (Domain empty), the sync fails with "Directory configuration incomplete.", `GSuiteDirectoryService.auth()` requires `domain` to be non-null regardless of `customer`.

### Screenshots or Videos

_No response_

### Additional Context

Root cause is in `libs/services/directory-services/gsuite-directory.service.ts`: `auth()` treats `this.dirConfig.domain == null` as fatal regardless of `customer`. Additionally, when both fields are set, both parameters are forwarded to the Directory API, which treats them as mutually exclusive and prefers `domain`, so which parameter takes effect is decided implicitly by the API rather than explicitly by the client.

Reproduced with the official 2026.6.1 CLI and a from-source build of current `main`; the desktop app shares the same service code. Platform-independent (verified on Windows).

I have a fix ready and would like to submit a PR: accept at least one of `domain` / `customer` in validation, and always send exactly one parameter, with `domain` keeping precedence when both are set, deliberately preserving the current effective sync scope of every existing configuration. The change is purely additive: customer-only configurations become possible; nothing changes for existing setups. Includes unit tests.

And also I am sorry for the heavy use of AI writing, I wanted this issue to be fixed as soon as possible, hopefully all of this is enough for getting a sense of the situation.

### Operating System

Windows

### Operating System Version

Windows 11

### Directory Service

Google Workspace

### Application Type

CLI (the bwdc command line application)

### Build Version

2026.6.1

### Issue Tracking Info

- [x] I understand that work is tracked outside of Github. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

watchingdogs:
## 🎟️ Tracking

ID: PM-40799
Link: https://bitwarden.atlassian.net/browse/PM-40799
Issue: https://github.com/bitwarden/directory-connector/issues/1193

## 📔 Objective

Allow a Google Workspace configuration with only `Customer ID` set, currently rejected as "Directory configuration incomplete", which leaves multi-domain accounts unable to sync all of their domains (see the linked issue for details).

- Validation now requires at least one of `domain` / `customer`.
- Exactly one query parameter is sent per request; `domain` keeps precedence when both are configured, so **existing configurations keep their current sync scope**, the change is purely additive.
- A whitespace-only `domain` no longer results in a parameterless request (previously surfaced as an obscure Directory API error).

Unit tests added for the validation and parameter selection. Verified against a live multi-domain tenant (8 users, 2 groups, one secondary domain):

| Config            | Before   | After              |
|-------------------|----------|--------------------|
| customer only     | rejected | 8 users, 2 groups  |
| domain + customer | 7 / 1    | 7 / 1 (unchanged)  |
| domain only       | 7 / 1    | 7 / 1 (unchanged)  |


eliykat:
Thanks for your PR @watchingdogs . Your issue write-up and proposed solution make sense to me, but I'm just double checking internally as I'm less familiar with the Google API specifically.

watchingdogs:
Hey @eliykat, do you have any updates? Is there anything I can assist with?

eliykat (approved):
The changes look good to me and internal feedback was positive. This apparently addresses a longstanding feature request - good work! I will pass to QA to test before merging. Note that this may be a couple of weeks depending on their capacity.

CC @BTreston - can we update any of our integration tests to cover this?

eliykat:
This has passed QA and has been marked ready to merge. I'm just confirming the status of our next DC release as to whether it makes this cut or the next one.

watchingdogs:
Glad to hear these. Would be awesome if it could make the current release.

eliykat (approved):
Thank you for the contribution! This is slated for the upcoming release this month.

PR #1195, opened 2026-07-23
[PM-40617] fix encoding issues during keytar migration

BTreston:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-40617

## 📔 Objective
Keytar wrote credentials via `CredWriteA` with UTF-8 blobs. desktop_core reads via `CredReadW` and interprets blobs as UTF-16 LE. The two formats are binary-incompatible. `NativeSecureStorageService.get()` silently returned `null` for any keytar written secrect, so the v3→5 migration's `secureStorageService.has(oldKey)` check always returned `false`. If you're running the migration on the version without keytar, then you cannot read the creds to migrate without converting the encoding. This was not properly implemented, therefore no windows credentials were ever correctly copied from {userId}_ldapPassword → secretLdap. You could observe this behavior by checking the entries in windows credential manager and see only the old userId prefixed keys.

The fix runs the v7 migration on windows only to re-encode the old secrets from the previous keys into the new key locations with the correct encoding. Does not overwrite keys written to the new locations for secrets written for installations already on state v6. Linux and MacOS only bump state version, no migration needed.

The rust migration code also needs to return a struct that can be mapped to a TS type, not a primitive bool. The struct has `pub error: Option<String>` property to report what is happening back to the client.

Note: legacy keys are still left in place from the v3→5 migration to allow for addressing potential issues like this until the migration issues have stabilized out in prod.

The new migration was tested locally on a windows 11 machine migrating from:

- `v2025.12.0` -> `v2026.6.0` -> development build. 

Confirmed that new credentials were migrated in windows credential manager and that the credentials are correctly read by the app and pulled into application state without errors.


src-gui/services/electron/electronStorage.service.ts
```diff
@@ -14,6 +14,20 @@ export class ElectronStorageService implements StorageService {
     if (!fs.existsSync(dir)) {
       NodeUtils.mkdirpSync(dir, "700");
     }
+
+    // On Windows, a previous version could write data.json as UTF-16 LE (BOM or null-prefixed).
+    // electron-store reads files as UTF-8 and throws "invalid nul value" when it encounters
+    // UTF-16 content. Detect and re-encode to UTF-8 before electron-store opens the file.
+    const dataFile = `${dir}/data.json`;
+    if (fs.existsSync(dataFile)) {
+      const raw = fs.readFileSync(dataFile);
```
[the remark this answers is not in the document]
BTreston: Not an issue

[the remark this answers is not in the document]
mandreko-bitwarden: @AikidoSec ignore: path is not attacker-controlled. Filename is a hardcoded literal (data.json) and the directory is `app.getPath("userData")`, an OS-defined path with no external input. Content is re-encoded in place and not returned anywhere.

libs/services/state-service/stateMigration.service.ts
```diff
@@ -58,9 +58,20 @@ export class StateMigrationService {
...
           break;
-        case StateVersion.Five:
+        case StateVersion.Five: {
+          // If activeUserId is still present the v3→5 credential migration previously failed
+          // (credentials were skipped due to the UTF-8/UTF-16 mismatch). Re-run it now so
+          // the secrets are copied before the v5→6 re-encoding step.
+          const hasLegacyData = (await this.storageService.get("activeUserId" as any)) != null;
+          if (hasLegacyData) {
+            await this.migrateStateFrom3To5();
+          }
```
eliykat: We should treat state migrations as immutable, otherwise we have more states to keep track of and more branching logic. e.g. here we will have some clients that ran the old `3to5` migration, some clients that run the new `3to5` migration, and we have to account for both states at each migration afterwards. And this distinction isn't clear for anyone reading the code later.

A simpler solution is to let everyone run the old migration, then fix everyone the same way in `6to7`.

libs/services/state-service/stateMigration.service.ts
```diff
@@ -169,13 +174,34 @@ export class StateMigrationService {
...
+          oldKey,
+          newKey,
+        );
+        if (!migrated) {
+          // Non-Windows: migrateKeytarPasswordAs is a no-op, so copy cross-platform.
+          const value = await this.secureStorageService.get<string>(oldKey);
+          if (value != null) {
             await this.secureStorageService.save(newKey, value);
           }
         }
```
eliykat, line 183: This split seems worse - now windows is handled entirely in rust, and others are a no-op passthrough that are handled with TS + rust. I suggest it would be simpler to do:

```ts
const value = await this.secureStorageService.getOldPassword<string>(oldKey); // this contains the diverging logic for correctly reading the value based on the platform
  await this.secureStorageService.save(newKey, value);
}
```

Thinking this through - this was the purpose of `get_password_keytar`, which we are already using indirectly in `5to6` via `migrateKeytarPassword`. Why didn't that address the encoding issue? Or if `5to6` was reading the new keys, ignoring the fact that we never migrated from the old keys, can't we just call `migrateKeytarPassword` with the old keys? (making some assumptions here, just trying to figure it out)

(As I said above I don't think we should be editing the old migration at all, but I'm assuming we need this logic somewhere)

libs/services/state-service/stateMigration.service.ts
```diff
@@ -169,13 +174,34 @@ export class StateMigrationService {
...
+        await this.secureStorageService.remove(oldKey);
```
eliykat: The PR description says we're not removing any values until the migration settles, which I agree with.

libs/services/state-service/stateMigration.service.ts
```diff
@@ -169,13 +174,34 @@ export class StateMigrationService {
...
+        `${clientId}_entraKey`,
+        SecureStorageKeys.entra,
+      );
+      if (!entraKeyMigrated) {
+        const value = await this.secureStorageService.get<string>(`${clientId}_entraKey`);
+        if (value != null) {
+          await this.secureStorageService.save(SecureStorageKeys.entra, value);
+        }
       }
+      await this.secureStorageService.remove(`${clientId}_entraKey`);
```
eliykat: Same issues as above

libs/services/state-service/stateMigration.service.ts
```diff
@@ -190,7 +216,7 @@ export class StateMigrationService {
           account.keys.apiKeyClientSecret,
         );
       }
-      if (account.tokens) {
+      if (account.tokens?.accessToken != null) {
```
eliykat: Is this required?

eliykat (changes requested):

eliykat:
I expect I'll have more feedback, but my main feedback from the initial review above is:
- don't change migrations that may have already run - treat them as immutable and fix everyone together later
- don't reimplement state migration logic in rust that we already have in TS - I'd prefer to use the helpers (such as `migrate_keytar_password`) that we already have available to us, assuming they work, and leave the account parsing/traversal to the TS layer where we're more comfortable with it. With these changes, different migrations are using different paths through the rust code and it becomes difficult to track.

BTreston:
@eliykat ~~I think this is ready. Most (all) of the complexity was from trying to rust it up. This is much better now.~~ I tested the build artifact generated by CI on a real windows PC (not a VM) and there is still an issue there. I may have paired away too much of what were valid fixes trying to simplify while testing using parallels... Sending back to draft until I fully iron it out. Apologies

libs/services/state-service/stateMigration.service.ts
```diff
@@ -265,6 +268,53 @@ export class StateMigrationService {
...
+        { old: `${clientId}_gsuitePrivateKey`, new: SecureStorageKeys.gsuite },
+        { old: `${clientId}_azureKey`, new: SecureStorageKeys.azure },
+        { old: `${clientId}_entraIdKey`, new: SecureStorageKeys.entra },
+        { old: `${clientId}_entraKey`, new: SecureStorageKeys.entra },
+        { old: `${clientId}_oktaToken`, new: SecureStorageKeys.okta },
+        { old: `${clientId}_oneLoginClientSecret`, new: SecureStorageKeys.oneLogin },
+        { old: `${clientId}_accessToken`, new: SecureStorageKeys.accessToken },
+        { old: `${clientId}_refreshToken`, new: SecureStorageKeys.refreshToken },
+        { old: `${clientId}_twoFactorToken`, new: SecureStorageKeys.twoFactorToken },
+      ];
```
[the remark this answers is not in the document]
BTreston: not an issue, these were not in keytar

BTreston:
@eliykat OK this should be ready for review now (real)

libs/services/state-service/stateMigration.service.ts
```diff
@@ -265,6 +268,59 @@ export class StateMigrationService {
...
+        { old: `${clientId}_entraIdKey`, new: SecureStorageKeys.entra },
+        { old: `${clientId}_entraKey`, new: SecureStorageKeys.entra },
+        { old: `${clientId}_oktaToken`, new: SecureStorageKeys.okta },
+        { old: `${clientId}_oneLoginClientSecret`, new: SecureStorageKeys.oneLogin },
+      ];
+
+      const written = new Set<string>();
+      for (const { old: oldKey, new: newKey } of oldSecretKeys) {
+        // Re-encode the UTF-8 keytar blob to UTF-16 in-place so desktop_core can read it.
+        await passwords.migrateKeytarPassword(SECURE_STORAGE_SERVICE_NAME, oldKey);
```
eliykat: We shouldn't touch the old values at all - not to delete them, but also not to re-encode them. As Claude [notes](https://github.com/bitwarden/directory-connector/pull/1195/changes#r3691906127), it's not idempotent, so if it re-runs for any reason it'll corrupt the data. Or if it just doesn't work for any reason, we will have corrupted the old data. Writing to a new location means we can always go back and try again no matter what.

Relatedly, the order of operations is unexpected:
1. re-encode oldKey value in place
2. skip if newKey value is already populated
3. overwrite newKey value

But if the newKey already exists and is OK, why touch oldKey at all?

I would expect:
1. skip if newKey value is already populated (I assume this shows a successful migration or new data that we shouldn't clobber? please confirm)
2. re-encode oldKey value to newKey location (single call to `migrateKeytarPassword`)

As far as I can tell, it's the same as the 5to6 migration but just targeting the oldKeys that were never migrated (re-encoding and migrating in one step) instead of newKeys (which were re-encoded in place, because already migrated).

eliykat (commented):
This is way easier to understand - great job simplifying it.

libs/services/state-service/stateMigration.service.ts
BTreston: I've added a separate read function. The flow should now be:

1. noop if not on windows or using plaintext secrets. (I'm assuming we don't want to extract secrets out of cred manager into plaintext based on a claude finding. We indicated this is what we wanted originally in state v3 but the param was unused so... If we don't want to I can remove it, but it makes sense to me to keep them separate)
2. skip newKey values that were written after v5-6 migration was run. This would be new data, not successfully migrated data. Successfully migrated data would only exist in non-windows installs where this migration is a noop anyway
3. immutable read of UTF-8 keytar data from old key, re-encoded as UTF-16 and returned to the client as a json blob
    - This is done to avoid re-encoding old key in place with `migrateKeytarPassword`
5. `JSON.parse` the response and save the now correctly encoded value returned from napi via `secureStorageService` under the newKey
6. bump state version.

I've also added more logging and some more tests.

BTreston: ~~After looking at the v5to6 migration again, I think the old keys may already have the correctly encoded values. I'll try updating this to just use `SecureStorageService.get()` on the old key and save it to the new key with `SecureStorageService.save()`. Its possible that's what's happening here, along with the need for return structs on the napi side. I'm attempting to test this on windows, but GH actions have been down for hours, so I cant generate a legit build artifact for final validation...~~

Nope, I was wrong.

eliykat (commented):
Holding off on reviewing further until you've [confirmed your hypothesis](https://github.com/bitwarden/directory-connector/pull/1195#discussion_r3731096789) of what's going on here, as it affects what the fix is. I've started a thread in Slack to discuss.

libs/services/state-service/stateMigration.service.ts
```diff
@@ -256,15 +262,81 @@ export class StateMigrationService {
...
+
+        const raw = await passwords.readKeytarPassword(SECURE_STORAGE_SERVICE_NAME, oldKey);
+        if (raw != null) {
+          // raw is the keytar blob verbatim (JSON.stringify'd by the old KeytarSecureStorageService).
+          // Parse it once so secureStorageService.save doesn't double-encode it.
+          let value: unknown;
+          try {
+            value = JSON.parse(raw);
+          } catch {
+            value = raw;
```
eliykat, line 329: What's the assumption here - if it's not parseable as json, assume it's already been parsed (or is otherwise not a json value), so just save it directly? I think that's reasonable, but worth adding an explanatory comment.

libs/services/state-service/stateMigration.service.ts
```diff
@@ -256,15 +262,81 @@ export class StateMigrationService {
...
+
+    const clientId = await this.storageService.get<string>("activeUserId");
+
+    if (clientId) {
+      const oldSecretKeys = [
+        { old: `${clientId}_ldapPassword`, new: SecureStorageKeys.ldap },
+        { old: `${clientId}_gsuitePrivateKey`, new: SecureStorageKeys.gsuite },
+        { old: `${clientId}_azureKey`, new: SecureStorageKeys.azure },
+        { old: `${clientId}_entraIdKey`, new: SecureStorageKeys.entra },
+        { old: `${clientId}_entraKey`, new: SecureStorageKeys.entra },
```
eliykat, line 304: Is this intentional - did we use both `entraIdKey` and `entraKey` at some stage?

native/src/migration.rs
```diff
@@ -45,10 +45,17 @@ fn get_password_keytar(service: &str, account: &str) -> Result<String> {
     Ok(String::from(password))
 }
 
+pub async fn read_keytar_password(service: &str, account: &str) -> Result<Option<String>> {
+    match get_password_keytar(service, account) {
+        Ok(v) => Ok(Some(v)),
+        Err(_) => Ok(None),
+    }
+}
```
eliykat, line 53: This method will never return an `Err`, so its signature could just be `Option<String>`. In that case it adds nothing, so you could just expose `get_password_keytar` directly.

That said, I don't understand why we're mapping `Err` -> `None`. Does `get_password_keytar` return an `Err` if no value is present? In other words - does `Err` really represent "no value found" or are we just swallowing errors here? If it's the latter, what _does_ `get_password_keytar` return if no value is found?

This was prompted by this [Claude comment](https://github.com/bitwarden/directory-connector/pull/1195#discussion_r3736530253), which I'm sympathetic to - logging real failures would be useful particularly given the trouble we've had here.

src-gui/app/services/services.module.ts
```diff
@@ -205,8 +205,12 @@ export const servicesProviders: (Provider | EnvironmentProviders)[] = [
...
   safeProvider({
     provide: StateMigrationService,
-    useClass: StateMigrationService,
-    deps: [StorageServiceAbstraction, SECURE_STORAGE],
+    useFactory: (
+      storageService: StorageServiceAbstraction,
+      secureStorageService: StorageServiceAbstraction,
+      logService: LogServiceAbstraction,
+    ) => new StateMigrationService(storageService, secureStorageService, logService),
+    deps: [StorageServiceAbstraction, SECURE_STORAGE, LogServiceAbstraction],
```
eliykat, line 213: The factory is hiding the missing parameter here. You need to pass in the `plaintextSecrets` value. Or is it a CLI-only option?

BTreston: It is CLI only. Default in ctor is `true`

libs/services/state-service/stateMigration.service.ts
BTreston: Yes, we did use both these keys previously. At this stage its probably fine to drop the old `entraKey` as it was pre v3, but I'd rather not right now.

native/src/migration.rs
BTreston: If theres no value found it still returns an `Err`. But there are different `Err` codes for different reasons. We can inspect it and handle it accordingly.

libs/services/state-service/stateMigration.service.ts
BTreston: added comment

eliykat (approved):
Great work! Please make sure QA have good instructions so they properly verify.

PR #1210, opened 2026-08-05
[PM-33583] AD User Filter Timeout issue

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33583

## 📔 Objective

Fixes issue in DC where timeout makes it seem like sync succeeds and brings back no results... Passing `timeLimit` to server allows the server to use it's own query policy for timeout instead.

**Before**
[directory-connector-timeout-issue.webm](https://github.com/user-attachments/assets/10715a3e-6fe6-4511-83d5-cc93c047fbb6)

**After**
[dc-no-longer-times-out-on-client-side.webm](https://github.com/user-attachments/assets/b56d2c66-cbae-4b67-b48a-2bcb696723ae)


BTreston (approved):

libs/services/directory-services/ldap-directory.service.ts
```diff
@@ -408,6 +408,11 @@ export class LdapDirectoryService implements IDirectoryService {
       filter: filter,
       scope: "sub",
       paged: this.dirConfig.pagedSearch,
+      // ldapts defaults timeLimit to 10 seconds, which the server enforces. Expensive but
+      // legitimate filters (e.g. LDAP_MATCHING_RULE_IN_CHAIN over nested groups) can exceed
+      // that and come back empty or as a timeLimitExceeded error. 0 defers to the server's
+      // own query policy instead.
+      timeLimit: 0,
```
BTreston: This seems ok on the surface, but I have no idea if it actually fixes the issue... We need to be able to guarantee this fixes the issue and provide test coverage for it.

BTreston (dismissed):

jrmccannon (approved):

PR #1212, opened 2026-08-11
[PM-41251] Disable nodeIntegration, add preload + IPC layer

BTreston:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-41251

## 📔 Objective
Disables Electron's `nodeIntegration` in the renderer process and establishes a secure IPC boundary. Migrates services that rely on node out of the renderer process and into the main process. These services are now called from the renderer process via the IPC layer.

Big changes:

- `nodeIntegration: false`  renderer no longer has direct Node.js access
- adds `preload.ts` which exposes a typed `window.ipc` API via electron's `contextBridge` so renderer can call its old services
- all services (auth, sync, secure storage, crypto, API) now instantiated in the main process
- adds IPC handlers registered via `ipcMain.handle`, which is a wrappper around the electron API w/ some extra error handling.
- added more platform utils for main process
- added `webpack.preload.mjs` config for preload bundle
- adds and updates context and documentation for claude.
- removes a lot of unused code left over the the jslib era of DC.

The renderer process should now be fully sandboxed. all privileged operations that require node run in main and are exposed only through whitelisted IPC channels.


src-gui/app/accounts/apiKey.component.ts
```diff
@@ -98,7 +96,7 @@ export class ApiKeyComponent {
     }
 
     try {
-      const promise = this.authService.logIn({ clientId, clientSecret });
+      const promise = ipc.auth.logIn({ clientId, clientSecret });
```
BTreston: @eliykat curious on your opinion on this pattern. I don't know if I am a fan of the IPC calls as they're done here, maybe we want an abstraction to sit between the IPC calls and these renderer call sites that we can DI into the components. Maybe each main process service that has an IPC channel, similar to what ended up in `src-gui/services/electron/electronRendererSecureStorage.service.ts`?

webpack.renderer.mjs
```diff
@@ -52,6 +52,7 @@ const common = {
   output: {
     filename: "[name].js",
     path: path.resolve(__dirname, "build"),
+    globalObject: "window",
```
BTreston, line 57: no `global` obj when node integration is disabled in renderer process.

BTreston (commented):
This all works, but its not perfect. It yanks everything that relies on node out of the renderer process in a pretty unceremonious manner. Theres probably room for improvement but this is the gist what we do in `clients`... Just looking for general feedback, this minimally resolves the linked issue. I can iterate on any async feedback once I am back from PTO so no rush here.

The bunch of comments are for future me, who will thank me when reading them in 2 weeks. 🤝

src-gui/services/electron/electronMainPlatformUtils.service.ts
BTreston, line 1: ♻️ We can still thin this out. There's leftovers from jslib that haven't been removed and are unused. But I'd consider it out of scope for now. Claude just copied them over 1:1 with the OG platform utils service.

src-gui/services/electron/electronPlatformUtils.service.ts
BTreston, line 1: Rename: `RendererPlatformUtils`

src-gui/services/electron/electronMainPlatformUtils.service.ts
BTreston, line 1: Rename: `MainPlatformUtils`

src-gui/services/electron/rendererMessaging.service.ts
BTreston, line 1: Rename: RendererMessaging

src-gui/services/electron/rendererSecureStorage.service.ts
BTreston, line 1: Rename: RendererSecureStorage

src-gui/services/electron/rendererStorage.service.ts
BTreston, line 1: Rename: RendererStorage

src-gui/main.ts
```diff
@@ -58,14 +72,120 @@ export class Main {
...
+      ipcMain.handle(channel, async (...args) => {
+        try {
+          return await (handler as (...a: typeof args) => any)(...args);
+        } catch (e: unknown) {
+          if (e instanceof Error) throw e;
+          const msg = (e as any)?.message ?? String(e);
+          throw new Error(msg);
+        }
+      });
+    };
```
BTreston: de-claude this

src-gui/utils.ts
BTreston, line 1: I think most of this file can go now we have these behind IPC calls. I think a lot are duplicates, verify and remove.

WEBPACK.md
BTreston, line 1: This is pretty useless info for a whole file. Can be comments in the webpack files.

BTreston:
While this addresses the ticket, this leaves DC in a rather fragile state. GUI build will bork if you accidentally pull in any node api in the renderer process, which is relatively easy to do accidentally (especially with `dc_native` being a straight up node binary that we manually vendor in) and end up with an arcane error message. Claude is decently good at following the runtime hierarchy based on what comes out of the build to find what node api is causing the issue, but its not a great dev-ex... I'm not sure if that's a big problem or if that's just how it is working with sandboxing in electron.

BTreston:
> While this addresses the ticket, this leaves DC in a rather fragile state. GUI build will bork if you accidentally pull in any node api in the renderer process, which is relatively easy to do accidentally (especially with `dc_native` being a straight up node binary that we manually vendor in) and end up with an arcane error message. Claude is decently good at following the runtime hierarchy based on what comes out of the build to find what node api is causing the issue, but its not a great dev-ex... I'm not sure if that's a big problem or if that's just how it is working with sandboxing in electron.

I have addressed this with a new linter rule preventing node API and electron imports in the renderer. I think that should take care of this concern

src-gui/services/electron/rendererI18n.service.ts
```diff
@@ -0,0 +1,14 @@
+import { I18nService as BaseI18nService } from "@/libs/services/baseI18n.service";
+
+export class RendererI18nService extends BaseI18nService {
+  constructor(systemLanguage: string, localesDirectory: string) {
+    super(systemLanguage, localesDirectory, async (formattedLocale: string) => {
+      const response = await fetch(`${localesDirectory}/${formattedLocale}/messages.json`);
```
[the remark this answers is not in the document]
BTreston: @AikidoSec ignore: false positive

src-gui/main.ts
```diff
@@ -110,31 +232,29 @@ export class Main {
...
-        this.messagingMain.init();
-        await this.updaterMain.init();
-        await this.trayMain.init(this.i18nService.t("bitwardenDirectoryConnector"));
-
-        if (!app.isDefaultProtocolClient("bwdc")) {
-          app.setAsDefaultProtocolClient("bwdc");
-        }
+    try {
+      await this.stateService.init();
+      await this.windowMain.init();
```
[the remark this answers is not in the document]
BTreston: This is a _potential_ race condition. It does not appear to happen for us, but [electron documentation](https://www.electronjs.org/docs/latest/api/app#event-ready) does support using `app.whenReady()` so this is valid.

BTreston:
@coltonhurst If you're available to give the just the IPC implementation here some feedback it would be much appreciated, as I've been told you have some experience with it.

src-gui/main.ts
```diff
@@ -17,6 +31,22 @@ import { DCCredentialStorageListener } from "./main/credential-storage-listener"
 import { MenuMain } from "./main/menu.main";
 import { MessagingMain } from "./main/messaging.main";
 
+// Normalize non-Error rejections to Error so they serialize correctly over IPC.
+function handle(channel: string, handler: Parameters<typeof ipcMain.handle>[1]) {
+  ipcMain.handle(channel, async (...args) => {
+    try {
+      return await handler(...args);
+    } catch (e: unknown) {
+      if (e instanceof Error) {
```
JaredScar, line 40: Unknown? Isn't it always going to be an Error if it's caught?

BTreston: There's no way TS can know that at compile time. You'll never see `catch (e: Error)` in TS code.

sven-bitwarden:
Direction seems sound and I haven't seen anything that I could identify as alarming. My only question: I assume the IPC layer entirely is managed by electron, so we have nothing to handle on our side in terms of transient failures (either for the connection as a whole, or on a per-message basis) between the renderer<->main process?

BTreston:
> Direction seems sound and I haven't seen anything that I could identify as alarming. My only question: I assume the IPC layer entirely is managed by electron, so we have nothing to handle on our side in terms of transient failures (either for the connection as a whole, or on a per-message basis) between the renderer<->main process?

Yes, electron manages the IPC wrt the messaging between renderer and main. The only thing we really have to concern ourselves with is the `handle()` function done in `src-gui/main.ts` to make sure non-error result objects (like failure messges from sync services) that come across the boundary are treated like errors.

jrmccannon (approved):
As much as I can tell it looks good.

